PR #11, opened 2018-07-13
Being able to refund units only from paid orders

GSadee:


spec/Checker/OrderRefundingAvailabilityCheckerSpec.php
```diff
@@ -0,0 +1,44 @@
...
+        OrderRepositoryInterface $orderRepository,
+        OrderInterface $order
+    ): void {
+        $orderRepository->findOneByNumber('00000007')->willReturn($order);
+        $order->getPaymentState()->willReturn(OrderPaymentStates::STATE_PAID);
+
+        $this->__invoke('00000007')->shouldReturn(true);
+    }
+
+    function it_returns_true_if_order_is_unpaid(
```
Zales0123: `returns_true` -> `returns_false`

spec/Checker/OrderRefundingAvailabilityCheckerSpec.php
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    function it_returns_true_if_order_is_paid(
+        OrderRepositoryInterface $orderRepository,
+        OrderInterface $order
+    ): void {
+        $orderRepository->findOneByNumber('00000007')->willReturn($order);
+        $order->getPaymentState()->willReturn(OrderPaymentStates::STATE_PAID);
+
+        $this->__invoke('00000007')->shouldReturn(true);
```
Zales0123: Why not `$this('00000007')->shouldReturn(true);`?

src/Action/OrderRefundsListAction.php
```diff
@@ -25,8 +30,24 @@ public function __construct(OrderRepositoryInterface $orderRepository, Environme
 
     public function __invoke(Request $request): Response
     {
+        /** @var OrderInterface $order */
         $order = $this->orderRepository->findOneByNumber($request->attributes->get('orderNumber'));
 
+        if ($order->getPaymentState() !== OrderPaymentStates::STATE_PAID) {
```
Zales0123: I wonder if this check is really necessary... Of course, it protects from accessing refunds list page for not paid order... one the other hand we already do not display the **refunds** button and even if somebody would go there by URL, their still would not be possible to refund anything, as we have also check it in the command handler... It just adds big complexity to this action for a relatively small benefit ⚖️

Zales0123: Or if we really need to check it, we can consider protecting it on the level of the security (by [expression](http://symfony.com/doc/current/security/access_control.html#securing-by-an-expression) or [voter](http://symfony.com/doc/current/security/voters.html))

src/CommandHandler/RefundUnitsHandler.php
```diff
@@ -18,21 +20,30 @@
...
         $this->refundCreator = $refundCreator;
         $this->refundedUnitTotalProvider = $refundedUnitTotalProvider;
+        $this->orderRefundingAvailabilityChecker = $orderRefundingAvailabilityChecker;
         $this->eventBus = $eventBus;
     }
 
     public function __invoke(RefundUnits $command): void
     {
+        if (!$this->orderRefundingAvailabilityChecker->__invoke($command->orderNumber())) {
+            throw OrderIsNotPaidException::withOrderNumber($command->orderNumber());
```
Zales0123: Maybe `OrderNotAvailableForRefunding` or sth similar? In fact we check if order is available for refunding, not if it's paid :)

tests/Behat/Context/Application/RefundingContext.php
```diff
@@ -13,6 +13,7 @@
 use Sylius\Component\Resource\Repository\RepositoryInterface;
 use Sylius\RefundPlugin\Command\RefundUnits;
 use Sylius\RefundPlugin\Entity\RefundInterface;
+use Sylius\RefundPlugin\Exception\OrderIsNotPaidException;
```
Zales0123: 😢

tests/Behat/Context/Ui/RefundingContext.php
```diff
@@ -31,7 +32,10 @@ public function __construct(
      */
     public function wantToRefundSomeUnitsOfOrder(string $orderNumber): void
     {
-        $this->orderRefundsPage->open(['orderNumber' => $orderNumber]);
+        try {
+            $this->orderRefundsPage->open(['orderNumber' => $orderNumber]);
+        } catch (UnexpectedPageException $exception) {
```
Zales0123: Do we really need to catch it if we do nothing about that? This way some exception would be just thrown in one of the next steps

src/Action/OrderRefundsListAction.php
Zales0123: Well, for me it still seems like a security, admin should not have **access** to order's refunds page 😄 

Nevertheless, we should for sure extract this to some service like `OrderRefundsListGuard`... Here it should be only calling a service and redirecting to referer if it fails/throws an exception.

Zales0123: ping @GSadee still relevant 😄

Zales0123:
Thank you Grzegorz! 🚜

PR #13, opened 2018-07-18
Refunding order shipping

Zales0123:
The only thing I'm not sure about is `RefundUnits` command, with which now we can also refund shipments. On one hand, it's quite confusing (the natural thing is we think about `order units`)... On the other hand, we can say that in a context of `Refund` everything is a **refund unit**, no matter is it a shipment or order unit. And I have no idea how can it be named better 😄 

> Note

For now, there is only a support for one order shipment - the reason why, is that there could be some troubles about shipments in Sylius (we have one shipping adjustment even for multiple shipments) and even though Sylius supports multiple shipments, it's base on single-shipment architecture right now.

zrzut ekranu 2018-07-19 o 08 49 05

### DIFF
--- a/features/refunding_order_shipping_cost.feature
+++ b/features/refunding_order_shipping_cost.feature
@@ -12,6 +12,7 @@ Feature: Refunding an order shipping cost
         And there is a customer "rick.sanchez@wubba-lubba-dub-dub.com" that placed an order "#00000022"
         And the customer bought 2 "Mr. Meeseeks T-Shirt" products
         And the customer chose "Galaxy Post" shipping method to "United States" with "Space money" payment
+        And the order "#00000022" is already paid
         And I am logged in as an administrator
 
     @ui @application
--- a/features/refunding_order_shipping_cost.feature
+++ b/features/refunding_order_shipping_cost.feature
@@ -1,0 +1,33 @@
+@refunds
+Feature: Refunding an order shipping cost
+    In order to give back money spent by Customer for order shipment
+    As an Administrator
+    I want to be able to refund an order shipping cost
+
+    Background:
+        Given the store operates on a single channel in "United States"
+        And the store has a product "Mr. Meeseeks T-Shirt" priced at "$10.00"
+        And the store has "Galaxy Post" shipping method with "$20.00" fee
+        And the store allows paying with "Space money"
+        And there is a customer "rick.sanchez@wubba-lubba-dub-dub.com" that placed an order "#00000022"
+        And the customer bought 2 "Mr. Meeseeks T-Shirt" products
+        And the customer chose "Galaxy Post" shipping method to "United States" with "Space money" payment
+        And I am logged in as an administrator
+
+    @ui @application
+    Scenario: Refunding an order shipment
+        When I want to refund some units of order "#00000022"
+        And I decide to refund order shipment
+        Then I should be notified that order shipment has been successfully refunded
+        And this order refunded total should be "$20.00"
+        And I should not be able to refund order shipment
+
+    @ui @application
+    Scenario: Refunding and order shipment along with order unit
+        When I want to refund some units of order "#00000022"
+        And I decide to refund order shipment and 1st "Mr. Meeseeks T-Shirt" product
+        Then I should be notified that order shippment has been successfully refunded
+        Then I should be notified that selected order units have been successfully refunded
+        And this order refunded total should be "$30.00"
+        And I should not be able to refund order shipment
+        And I should not be able to refund 1st unit with product "Mr. Meeseeks T-Shirt"
--- a/features/refunding_order_shipping_cost.feature
+++ b/features/refunding_order_shipping_cost.feature
@@ -18,7 +18,6 @@ Feature: Refunding an order shipping cost
     Scenario: Refunding an order shipment
         When I want to refund some units of order "#00000022"
         And I decide to refund order shipment
-        Then I should be notified that order shipment has been successfully refunded
         And this order refunded total should be "$20.00"
         And I should not be able to refund order shipment
 
@@ -26,7 +25,6 @@ Feature: Refunding an order shipping cost
     Scenario: Refunding and order shipment along with order unit
         When I want to refund some units of order "#00000022"
         And I decide to refund order shipment and 1st "Mr. Meeseeks T-Shirt" product
-        Then I should be notified that order shipment has been successfully refunded
         Then I should be notified that selected order units have been successfully refunded
         And this order refunded total should be "$30.00"
         And I should not be able to refund order shipment
--- a/features/refunding_order_shipping_cost.feature
+++ b/features/refunding_order_shipping_cost.feature
@@ -22,7 +22,7 @@ Feature: Refunding an order shipping cost
         And this order refunded total should be "$20.00"
         And I should not be able to refund order shipment
 
-    @application
+    @application @todo
     Scenario: Refunding and order shipment along with order unit
         When I want to refund some units of order "#00000022"
         And I decide to refund order shipment and 1st "Mr. Meeseeks T-Shirt" product
--- a/features/refunding_order_shipping_cost.feature
+++ b/features/refunding_order_shipping_cost.feature
@@ -14,7 +14,7 @@ Feature: Refunding an order shipping cost
         And the customer chose "Galaxy Post" shipping method to "United States" with "Space money" payment
         And I am logged in as an administrator
 
-    @ui @application
+    @application
     Scenario: Refunding an order shipment
         When I want to refund some units of order "#00000022"
         And I decide to refund order shipment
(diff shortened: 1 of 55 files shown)


src/Event/UnitsRefunded.php
```diff
@@ -11,21 +11,13 @@
...
-    private $orderNumber;
-
-    /** @var iterable */
-    private $unitIds;
-
-    /** @var int */
-    private $amount;
-
-    public function __construct(string $orderNumber, iterable $unitIds, int $amount)
+    public function __construct(string $orderNumber, iterable $unitIds, iterable $shipmentIds, int $amount)
```
bartoszpietrzak1994: To be honest, when it comes to class naming, the difference between `UnitRefunded` and `UnitsRefunded` is so slight that it took me a few moments to figure out what's going on. WDYT about `OrderUnitsRefunded` or something that will let everyone easily distinguish between these two events?

src/Request/RefundUnitsRequest.php
```diff
@@ -6,19 +6,23 @@
...
 {
     public static function getCommand(Request $request): RefundUnits
     {
-        Assert::notNull($request->request->get('sylius_refund_units'), 'sylius_refund.at_least_one_unit_should_be_selected_to_refund');
+        if ($request->request->get('sylius_refund_units') === null && $request->request->get('sylius_refund_shipments') === null) {
+            throw new \InvalidArgumentException('sylius_refund.at_least_one_unit_should_be_selected_to_refund');
+        }
 
         return new RefundUnits(
             $request->attributes->get('orderNumber'),
```
bartoszpietrzak1994: To be honest, I find the code below hardly readable :D

tests/Behat/Context/Application/RefundingContext.php
```diff
@@ -84,7 +103,23 @@ public function shouldNotBeAbleToRefundUnitWithProduct(int $unitNumber, string $
         $unit = $this->getOrderUnit($unitNumber, $productName);
 
         try {
-            $this->commandBus->dispatch(new RefundUnits($this->order->getNumber(), [$unit->getId()]));
+            $this->commandBus->dispatch(new RefundUnits($this->order->getNumber(), [$unit->getId()], []));
+        } catch (CommandDispatchException $exception) {
+            return;
```
bartoszpietrzak1994: Does it mean that this step will pass if `CommandDispatchException` is thrown?

tests/Behat/Context/Application/RefundingContext.php
```diff
@@ -84,7 +103,23 @@ public function shouldNotBeAbleToRefundUnitWithProduct(int $unitNumber, string $
...
+     * @Then I should not be able to refund order shipment
+     */
+    public function shouldNotBeAbleToRefundOrderShipment(): void
+    {
+        $shippingAdjustment = $this->order->getAdjustments(AdjustmentInterface::SHIPPING_ADJUSTMENT)->first();
+
+        try {
+            $this->commandBus->dispatch(new RefundUnits($this->order->getNumber(), [], [$shippingAdjustment->getId()]));
         } catch (CommandDispatchException $exception) {
             return;
```
bartoszpietrzak1994: Does it mean that this step will pass if `CommandDispatchException` is thrown?

spec/CommandHandler/RefundUnitsHandlerSpec.php
```diff
@@ -9,50 +9,47 @@
...
 
         $eventBus->dispatch(Argument::that(function (UnitsRefunded $event): bool {
             return
                 $event->orderNumber() === '000222' &&
                 $event->unitIds() === [1, 3] &&
-                $event->amount() === 1500
-            ;
+                $event->shipmentIds() === [3, 4] &&
+                $event->amount() === 7000
+                ;
```
bartoszpietrzak1994: Formatting

tests/Behat/Context/Application/RefundingContext.php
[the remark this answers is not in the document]
Zales0123: Yes, because it's the desired behaviour 😄 Step is named `I should not be able to refund` so we expect the exception to be thrown :)

tests/Behat/Context/Application/RefundingContext.php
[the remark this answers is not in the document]
Zales0123: The same here

tests/Behat/Context/Application/RefundingContext.php
```diff
@@ -109,7 +144,7 @@ public function shouldBeAbleToRefundUnitWithProduct(int $unitNumber, string $pro
     /**
      * @Then I should be notified that selected order units have been successfully refunded
      */
-    public function shouldBeNotifiedThatSelectedOrderUnitsHaveBeenSuccessfullyRefunded(): void
+    public function notificationSteps(): void
     {
         // intentionally left blank - not relevant in application scope
```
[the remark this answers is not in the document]
Zales0123: Yes, but for now we use the same scenario for testing the `@application` and `@ui` - therefore without this step, it won't be passing for the `@application` tag

tests/Behat/Context/Application/RefundingContext.php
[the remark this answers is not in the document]
bartoszpietrzak1994: Yup, my bad!

src/Event/UnitsRefunded.php
bartoszpietrzak1994: For me the perfect naming would be `RefundLineItems` as both order items and shipments are LineItems in the context of Invoice and the result of a refund is Credit Memo creation. However, since we want to separate RefundPlugin and InvoicingPlugin from each other, I think your idea for renaming these events is just fine and acceptable :)

Zales0123: cc @GSadee @pamil I would like to see your opinion 😄

GSadee: For me, `OrderUnitRefunded` is even more confusing than `UnitRefunded`, because we don't have an `OrderUnit`, but `OrderItemUnit` 🎉 

And `RefundLineItems` is not a best solution imo, because we don't have this concept in this plugin.

To be honest, none of the proposed changes convinces me, I would be for leaving that as it is now.

src/Command/RefundUnits.php
```diff
@@ -11,12 +11,13 @@
 {
     use PayloadTrait;
 
-    public function __construct(string $orderNumber, array $refundedUnitIds)
+    public function __construct(string $orderNumber, array $refundedUnitIds, array $refundedShipmentIds)
```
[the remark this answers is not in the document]
GSadee: IMO `$unitIds` or `$orderItemUnitIds` instead of `$orderUnitIds`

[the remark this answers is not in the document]
Zales0123: `$unitIds` and `$shipmentIds` looks fair for me 👍

src/Event/UnitsRefunded.php
Zales0123: Ok, so to not make it longer - I would vote for merging it as it is and I'll open an issue about that naming, ok? @bartoszpietrzak1994 @GSadee

pamil: I think it's okay right now as it is. They only thing I would change in this line of code is `iterable` to `array` (as event payload should be serializable, and `iterable` here is way too general).

features/refunding_order_shipping_cost.feature
```diff
@@ -0,0 +1,32 @@
...
+        And the store has a product "Mr. Meeseeks T-Shirt" priced at "$10.00"
+        And the store has "Galaxy Post" shipping method with "$20.00" fee
+        And the store allows paying with "Space money"
+        And there is a customer "rick.sanchez@wubba-lubba-dub-dub.com" that placed an order "#00000022"
+        And the customer bought 2 "Mr. Meeseeks T-Shirt" products
+        And the customer chose "Galaxy Post" shipping method to "United States" with "Space money" payment
+        And the order "#00000022" is already paid
+        And I am logged in as an administrator
+
+    @ui @application
```
GSadee: Could we have tags in the first line of this file?

spec/CommandHandler/RefundUnitsHandlerSpec.php
```diff
@@ -9,50 +9,47 @@
...
-        RefundedUnitTotalProviderInterface $refundedUnitTotalProvider,
+        RefunderInterface $orderUnitsRefunder,
+        RefunderInterface $orderShipmentsRefunder,
         OrderRefundingAvailabilityCheckerInterface $orderRefundingAvailabilityChecker,
         EventBus $eventBus
     ): void {
         $this->beConstructedWith(
-            $refundCreator,
-            $refundedUnitTotalProvider,
+            $orderUnitsRefunder,
```
GSadee: `$orderItemUnitsRefunder`?

spec/Entity/RefundSpec.php
```diff
@@ -6,29 +6,37 @@
 
 use PhpSpec\ObjectBehavior;
 use Sylius\RefundPlugin\Entity\RefundInterface;
+use Sylius\RefundPlugin\Model\RefundType;
 
 final class RefundSpec extends ObjectBehavior
 {
+    function let()
```
GSadee: Missing return type

src/Model/RefundType.php
```diff
@@ -0,0 +1,40 @@
...
+
+declare(strict_types=1);
+
+namespace Sylius\RefundPlugin\Model;
+
+use Sylius\RefundPlugin\Exception\RefundTypeNotResolved;
+
+final class RefundType
+{
+    public const ORDER_UNIT = 'order_unit';
```
GSadee: IMO `ORDER_ITEM_UNIT` or just `UNIT`?

features/refunding_order_shipping_cost.feature
Zales0123: To be honest, I really hate how scenarios look with no tags above them :/ And there are fewer things to change if some only-ui or only-application scenario would be added to this file in the future

GSadee: 😢

PR #18, opened 2018-07-25
Missing credit memo migration

bartoszpietrzak1994:


migrations/Version20180725110313.php
```diff
@@ -0,0 +1,31 @@
+<?php declare(strict_types=1);
```
GSadee: `declare(strict_types=1);` to the new line

Zales0123: Move it two line below
```php
<?php

declare(strict_types=1);
```

migrations/Version20180725110313.php
```diff
@@ -0,0 +1,31 @@
...
+
+namespace Application\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
```
Zales0123: We can remove this comment

migrations/Version20180725110313.php
```diff
@@ -0,0 +1,31 @@
...
+final class Version20180725110313 extends AbstractMigration
+{
+    public function up(Schema $schema) : void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'mysql', 'Migration can only be executed safely on \'sqlite\'.');
```
Zales0123: To generate the migration properly, the best would be to configure mysql for test application in plugin and then generate it - there are some significant differences between mysql and sqlite migrations :)

PR #29, opened 2018-08-03
Send credit memo to customer automatically

Zales0123:


src/File/FileManagerInterface.php
```diff
@@ -0,0 +1,14 @@
...
+
+namespace Sylius\RefundPlugin\File;
+
+interface FileManagerInterface
+{
+    public function createWithContent(string $fileName, string $content): void;
+
+    public function remove(string $fileName): void;
+
+    public function getBaseDirectory(): string;
```
pamil: This leaks some implementation logic (the concept of a base directory, where files are stored directly under). What do you think about changing it to `public function realpath(string $fileName): string`, which would return an absolute, canonicalised path to the file?

src/Listener/CreditMemoGeneratedEventListener.php
```diff
@@ -0,0 +1,55 @@
...
+        /** @var CreditMemoInterface|null $creditMemo */
+        $creditMemo = $this->creditMemoRepository->findOneBy(['number' => $event->number()]);
+        if ($creditMemo === null) {
+            throw CreditMemoNotFound::withNumber($event->number());
+        }
+
+        /** @var OrderInterface|null $order */
+        $order = $this->orderRepository->findOneByNumber($event->orderNumber());
+        if ($order === null) {
+            throw OrderNotFound::withOrderNumber($event->orderNumber());
```
pamil: `withOrderNumber` -> `withNumber`? It would be more consistent with the `CreditMemoNotFound` exception.

src/Listener/CreditMemoGeneratedEventListener.php
```diff
@@ -0,0 +1,55 @@
...
+        }
+
+        assert($order->getCustomer() !== null);
```
pamil: Could we change that to a Webmozart/Assert call or an if? `assert` functions can be actually disabled in the runtime and PHP docs states:

> Assertions should be used as a debugging feature only. You may use them for sanity-checks that test for conditions that should always be TRUE and that indicate some programming errors if not or to check for the presence of certain features like extension functions or certain system limits and features.

http://php.net/manual/pl/function.assert.php

src/Resources/config/services.xml
```diff
@@ -71,5 +71,15 @@
...
         </service>
+
+        <service id="Sylius\RefundPlugin\Sender\CreditMemoEmailSender">
+            <argument type="service" id="Sylius\RefundPlugin\Generator\CreditMemoPdfFileGenerator" />
+            <argument type="service" id="sylius.email_sender" />
+            <argument type="service" id="Sylius\RefundPlugin\File\FileManager" />
+        </service>
+
+        <service id="Sylius\RefundPlugin\File\FileManager">
+            <argument type="string">%kernel.project_dir%/web/media/credit-memos/</argument>
```
pamil: Storing credits memo here is a security leak:
 - it is accessible by HTTP
 - filenames are generated based on the credit memo number, which is sequential

Meaning that you can download all credit memos in a blink of an eye.

We should use fully random, long names (which again can be exploited, if HTTP server is not configured correctly and lists all files in a directory when there's no index file, but it's still a lot better). The best solution would be to have these files outside web root and serve them through a PHP controller, where we can implement additional security rules (see `BinaryFileResponse`).

src/Listener/CreditMemoGeneratedEventListener.php
Zales0123: Sure!

src/Resources/config/services.xml
Zales0123: @pamil do you think that saving them in a temporary directory with `sys_get_temp_dir()` would solve the problem for now (as it's done in `InvoicingPlugin`? I agree that these files should have some random names generated, it could be solved for both plugins in the nearest future.

pamil:
Thank you, Matthias! 🎉

PR #31, opened 2018-08-07
I want to browse and download credit memo

bartoszpietrzak1994:
screen shot 2018-08-07 at 12 25 36


features/downloading_credit_memo_on_customer_order_view.feature
```diff
@@ -0,0 +1,25 @@
+@customer_browsing_credit_memos
+Feature: Downloading credit memo on customer order view
+    In order to have access to all credit memos related to order refunds
```
Zales0123: `In order to save generated credit memos for future usage`?

features/downloading_credit_memo_on_customer_order_view.feature
```diff
@@ -0,0 +1,25 @@
...
+
+    Background:
+        Given the store operates on a single channel in "United States"
+        And the store has a product "Angel T-Shirt"
+        And the store ships everywhere for free
+        And the store allows paying with "Cash on Delivery"
+        And I am a logged in customer
+        And I placed an order "#00000666"
+        And I bought a single "Angel T-Shirt"
+        And I addressed it to "Lucifer Morningstar", "Seaside Fwy", "90802" "Los Angeles" in the "United States"
```
Zales0123: `with identical billing address` :)

tests/Behat/Page/Shop/Order/ShowPage.php
```diff
@@ -12,4 +13,38 @@ public function countCreditMemos(): int
...
         return count($this->getDocument()->findAll('css', '#credit-memos tbody tr'));
     }
+
+    public function downloadFirstCreditMemo(): void
+    {
+        $creditMemo = $this->getFirstCreditMemo();
+        $creditMemo->clickLink('Download');
+    }
+
+    public function isPdfFileDownloaded(): bool
```
Zales0123: It should not be implemented here, we have `Tests\Sylius\RefundPlugin\Behat\Context\Element\PdfDownloadElement` (which in fact should be moved to `Tests\Sylius\RefundPlugin\Behat\Element\PdfDownloadElement`)

tests/Behat/Page/Shop/Order/ShowPage.php
```diff
@@ -12,4 +13,38 @@ public function countCreditMemos(): int
     {
         return count($this->getDocument()->findAll('css', '#credit-memos tbody tr'));
     }
+
+    public function downloadFirstCreditMemo(): void
```
Zales0123: What about something more flexible like `downloadCreditMemo(int $index): void`?

src/Resources/config/shop_routing.yml
```diff
@@ -8,3 +8,9 @@ sylius_refund_plugin_shop_order_credit_memos_partial:
             repository:
                 method: findByOrderNumber
                 arguments: $orderNumber
+
+sylius_refund_shop_credit_memo_download:
+    path: /credit-memos/{id}/download
+    methods: [GET]
+    defaults:
+        _controller: Sylius\RefundPlugin\Action\DownloadCreditMemoAction
```
Zales0123: I think this is a possible security issue - we're using the same action for Admin panel and Customer account, therefore it's possible for a Customer to access any credit memo in the system :( We should create a new action that would check the possibility to download credit memo... There is not so much logic to duplicate :)

src/Action/Shop/DownloadCreditMemoAction.php
```diff
@@ -0,0 +1,70 @@
...
+
+        /** @var CustomerInterface $orderCustomer */
+        $orderCustomer = $order->getCustomer();
+
+        /** @var CustomerInterface $customer */
+        $customer = $this->customerContext->getCustomer();
+
+        if ($orderCustomer->getId() !== $customer->getId()) {
+            return new Response(Response::HTTP_UNAUTHORIZED);
+        }
```
Zales0123: This whole logic should be extracted to separate service, that would throw an exception if a customer is not permitted to perform an action

src/Action/Shop/DownloadCreditMemoAction.php
```diff
@@ -0,0 +1,70 @@
...
+
+        $creditMemoPdfFile = $this->creditMemoPdfFileGenerator->generate($id);
+
+        $response = new Response($creditMemoPdfFile->content(), Response::HTTP_OK, ['Content-Type' => 'application/pdf']);
+        $response->headers->add([
+            'Content-Disposition' => $response->headers->makeDisposition('attachment', $creditMemoPdfFile->filename()),
+        ]);
+
+        return $response;
```
Zales0123: I think we can even extract this one to something like `CreditMemoFileResponseBuilder` as it's used in two actions :)

src/Checker/CreditMemoCustomerRelationChecker.php
```diff
@@ -0,0 +1,52 @@
...
+        /** @var OrderInterface $order */
+        $order = $this->orderRepository->findOneByNumber($creditMemo->getOrderNumber());
+
+        /** @var CustomerInterface $orderCustomer */
+        $orderCustomer = $order->getCustomer();
+
+        /** @var CustomerInterface $customer */
+        $customer = $this->customerContext->getCustomer();
+
+        Assert::eq($orderCustomer->getId(), $customer->getId());
```
Zales0123: We can think about some more sophisticated exception like `CreditMemoNotAccessible` or sth similar 😄

src/Checker/CreditMemoCustomerRelationChecker.php
```diff
@@ -0,0 +1,54 @@
...
+
+    public function check(string $creditMemoId): void
+    {
+        /** @var CreditMemoInterface $creditMemo */
+        $creditMemo = $this->creditMemoRepository->find($creditMemoId);
+
+        /** @var OrderInterface $order */
+        $order = $this->orderRepository->findOneByNumber($creditMemo->getOrderNumber());
+
+        /** @var CustomerInterface $orderCustomer */
```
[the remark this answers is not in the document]
bartoszpietrzak1994: Unfortunately, static code analysis does not pass without these type hints.

PR #42, opened 2018-08-21
Comment on credit memo

Zales0123:
zrzut ekranu 2018-08-21 o 13 20 21
zrzut ekranu 2018-08-21 o 13 29 35

### DIFF
--- a/src/Resources/config/doctrine/CreditMemo.orm.xml
+++ b/src/Resources/config/doctrine/CreditMemo.orm.xml
@@ -18,6 +18,7 @@
         <field name="units" type="json" />
         <field name="currencyCode" />
         <field name="localeCode" />
+        <field name="comment" />
     </entity>
 
 </doctrine-mapping>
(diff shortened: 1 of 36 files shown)


src/Resources/config/doctrine/CreditMemo.orm.xml
```diff
@@ -18,6 +18,7 @@
         <field name="units" type="json" />
         <field name="currencyCode" />
         <field name="localeCode" />
+        <field name="comment" />
```
GSadee: Is this field required?

GSadee: For me `comment` is an optional field, so it could be nullable, but I'm not fully convinced by any of these approaches, so it could stay at it is right now. 
Wdyt @bartoszpietrzak1994 ?

bartoszpietrzak1994: Making this field nullable may lead to state where you have both NULL and empty comments in the database. I would stay with the current solution :D

Zales0123: Agreed 👍

Zales0123: @bartoszpietrzak1994 Should be fixed now

PR #46, opened 2018-08-24
Filtering credit memos

Zales0123:
Based on https://github.com/Sylius/RefundPlugin/pull/45

zrzut ekranu 2018-08-24 o 15 21 07

### DIFF
--- a/features/filtering_credit_memos.feature
+++ b/features/filtering_credit_memos.feature
@@ -1,0 +1,38 @@
+@refunds
+Feature: Filtering credit memos
+    In order to browse credit memos more efficiently
+    As an Administrator
+    I want to be able to filter credit memos list
+
+    Background:
+        Given the store operates on a channel named "Web-US" in "USD" currency
+        And the store operates on a channel named "Web-UK" in "GBP" currency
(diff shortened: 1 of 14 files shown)


GSadee:
Could you rebase?

bartoszpietrzak1994:
Could you add a migration file related to CreditMemoChannel entity in order not to open an additional PR?

Zales0123:
It should be indeed added in the previous PR. Nevertheless, it's fixed now :)

Zales0123:
I was wondering about that, but this is a built-in functionality of `GridBundle`, this filtering functionality is already tested in a bundle and in **Sylius** itself. Even a custom filter is somehow tested, as it fulfills the interface (but this is the place where we could add some test, eventually). 🏹

GSadee:
It won't be a big benefit, so I won't insist on that, but IMO we should cover such tests with behat.

Zales0123:
Ok, I would vote to merge it to have it already in plugin (with no big harm) and think about most effective way to test it, wdyt?

GSadee:
Thank you Matt ⛺️

PR #47, opened 2018-08-27
Order unit partial refund preparation

Zales0123:
To avoid long-as-hell PR, I open this one with all code base prepared for the upcoming feature of partial refunds. It contains:

- partial refund scenarios (with `@todo` tag)
- `UnitRefund` value object (I'm open for naming discussion), that is used instead of plain unit id everywhere where it's needed
- default full price in `UnitRefund` VO (for both UI and application)

The next step would be allowing passing partial refund from UI and test it in both application and UI.

This is also a lesson, that we should use VO whenever we use some values that have a business sense 😄it would save us a lot of work


GSadee:
Thanks Mati 🚋

PR #53, opened 2018-09-05
Partial shipment refunds

Zales0123:
zrzut ekranu 2018-09-05 o 09 55 23


tests/Application/app/config/config_test.yml
```diff
@@ -5,6 +5,6 @@ doctrine:
     dbal:
         path: "%kernel.project_dir%/var/db_test.sql"
 
-knp_snappy:
-    pdf:
-        binary: "%kernel.project_dir%/app/etc/wkhtmltopdf"
+#knp_snappy:
+#    pdf:
+#        binary: "%kernel.project_dir%/app/etc/wkhtmltopdf"
```
bartoszpietrzak1994: To be removed

tests/Behat/Context/Application/RefundingContext.php
```diff
@@ -115,8 +116,27 @@ public function decideToRefundPartFromProductWithPayment(
...
+    public function decideToRefundPartOfOrderShipment(int $amount, PaymentMethodInterface $paymentMethod): void
+    {
+        $shippingAdjustment = $this->order->getAdjustments(AdjustmentInterface::SHIPPING_ADJUSTMENT)->first();
+
+        try {
+            $this->commandBus->dispatch(new RefundUnits(
+                $this->order->getNumber(), [], [new ShipmentRefund($shippingAdjustment->getId(), $amount)], $paymentMethod->getId(), ''
+            ));
+        } catch (CommandDispatchException $exception) {
+            return;
```
bartoszpietrzak1994: ```throw new \Exception('RefundUnits command should not fail');```

wdyt?

tests/Behat/Context/Ui/RefundingContext.php
```diff
@@ -203,6 +214,14 @@ public function shouldNotBeAbleToRefundUnitWithProduct(int $unitNumber, string $
         Assert::false($this->orderRefundsPage->isUnitWithProductAvailableToRefund($productName, $unitNumber-1));
     }
 
+    /**
+     * @Then I should still be able to refund order shipment with :paymentMethodName payment
+     */
+    public function shouldStillBeAbleToRefundOrderShipment(): void
+    {
+        Assert::true($this->orderRefundsPage->isOrderShipmentAvailableToRefund());
```
bartoszpietrzak1994: Since :paymentMethodName parameter is not used, maybe we can get rid of it?

Zales0123: We cannot, as the step name is the same as in `@application` test, where this payment method is relevant :/

tests/Behat/Context/Application/RefundingContext.php
Zales0123: I this specific case we want to silence the error, to check if anything changed later... but I think this step could be named `I try to refund` instead of `I decide to refund`, to avoid confusion... wdyt?

PR #55, opened 2018-09-10
Adding check if there are shippments

mamazu:
If there is no shipment the template rendering crashes because ` order.getAdjustments('shipping').first()` returns `false`


Zales0123:
Thank you @mamazu!

PR #62, opened 2018-09-17
Too long comment validation

Zales0123:
Before:

zrzut ekranu 2018-09-17 o 10 54 09

After:

zrzut ekranu 2018-09-17 o 15 13 00

### DIFF
--- a/src/Validator/RefundUnitsCommandValidator.php
+++ b/src/Validator/RefundUnitsCommandValidator.php
@@ -9,7 +9,7 @@ use Sylius\RefundPlugin\Command\RefundUnits;
 use Sylius\RefundPlugin\Exception\CreditMemoCommentTooLongException;
 use Sylius\RefundPlugin\Exception\OrderNotAvailableForRefundingException;
 
-final class RefundUnitsCommandValidator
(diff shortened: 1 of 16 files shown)


src/Validator/RefundUnitsCommandValidator.php
```diff
@@ -0,0 +1,37 @@
...
+    {
+        $this->orderRefundingAvailabilityChecker = $orderRefundingAvailabilityChecker;
+    }
+
+    /**
+     * @throws OrderNotAvailableForRefundingException|CreditMemoCommentTooLongException
+     */
+    public function validate(RefundUnits $command): void
+    {
+        if (!$this->orderRefundingAvailabilityChecker->__invoke($command->orderNumber())) {
```
lchrusciel: Probably could be written as `($this->orderRefundingAvailabilityChecker)($command->orderNumber())`

Zales0123:
Right now I'm thinking... does it really matter how long is this comment from our domain point of view? :D I think no, we can just change `comment` type to `text` to avoid constraints on a database and that's it 🚀

Zales0123:
Closing in favor of #64

PR #69, opened 2018-09-24
Change "sylius-plugin" to "symfony-bundle"

Zales0123:
In order to make this plugin work with SymfonyFlex (see https://github.com/symfony/recipes-contrib/pull/508), I propose a temporary (or maybe not) solution to change `type` in `composer.json` to `symfony-bundle` and put `sylius-plugin` as a keyword (which is also searchable by packagist). If it works, we should consider doing this for all Sylius plugins.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,6 +1,7 @@
 {
     "name": "sylius/refund-plugin",
-    "type": "sylius-plugin",
+    "type": "symfony-bundle",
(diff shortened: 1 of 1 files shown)


fabpot:
I've just added support for Sylius plugins in the Symfony Flex server, so this could be reverted now.

PR #74, opened 2018-09-27
Removed channel resolver in OrderRefundsListAction and used order's channel

### LINKED ISSUE #56: Refunds are based on the channel context
The refund process gets the possible payment methods for the refund from the current channel when it should be the channel of the order.

mamazu:
Fixes #56


Zales0123:
Good catch, @mamazu, thank you!

PR #84, opened 2018-10-02
Change partial refunds inputs UI + fix UI validation

Zales0123:
Before:

zrzut ekranu 2018-10-02 o 11 15 32

After:

zrzut ekranu 2018-10-02 o 11 15 44


pamil:
💪

PR #86, opened 2018-10-02
Support only for offline payment methods

Zales0123:
See PS in https://github.com/Sylius/RefundPlugin/pull/85 description

### DIFF
--- a/src/Action/Admin/RefundUnitsAction.php
+++ b/src/Action/Admin/RefundUnitsAction.php
@@ -12,6 +12,7 @@ use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpFoundation\Session\Session;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
+use Webmozart\Assert\Assert;
 
 final class RefundUnitsAction
 {
@@ -46,6 +47,7 @@ final class RefundUnitsAction
 
             $this->session->getFlashBag()->add('success', 'sylius_refund.units_successfully_refunded');
         } catch (CommandDispatchException $exception) {
+            Assert::notNull($exception->getPrevious());
             $this->session->getFlashBag()->add('error', $exception->getPrevious()->getMessage());
         } catch (\InvalidArgumentException $exception) {
             $this->session->getFlashBag()->add('error', $exception->getMessage());
--- a/src/Entity/CreditMemoUnit.php
+++ b/src/Entity/CreditMemoUnit.php
@@ -40,11 +40,13 @@ class CreditMemoUnit implements CreditMemoUnitInterface
 
     public function serialize(): string
     {
-        return json_encode([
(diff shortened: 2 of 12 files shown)


src/Entity/CreditMemoUnit.php
```diff
@@ -40,11 +40,13 @@ public function getTaxesTotal(): int
...
     public function serialize(): string
     {
-        return json_encode([
+        $serialized = json_encode([
             'product_name' => $this->productName,
             'total' => $this->total,
             'taxes_total' => $this->taxesTotal,
         ]);
+
+        return ($serialized !== false) ? $serialized : '';
```
pamil: I think we should throw an exception instead of silently returning an empty string.

Zales0123: Agreed 👍

src/Provider/OfflineRefundPaymentMethodsProvider.php
```diff
@@ -0,0 +1,36 @@
...
+
+        /** @var PaymentMethodInterface $paymentMethod */
+        foreach ($paymentMethods as $key => $paymentMethod) {
+            Assert::notNull($paymentMethod->getGatewayConfig());
+            if ($paymentMethod->getGatewayConfig()->getFactoryName() !== 'offline') {
+                unset($paymentMethods[$key]);
+            }
+        }
+
+        return $paymentMethods;
```
pamil: Not a big one, but you can get rid of that `unset` and phpdoc by using `array_filter`:

```php
return array_filter(
  $this->paymentMethodRepository->findEnabledForChannel($channel),
  function (PaymentMethodInterface $paymentMethod): bool {
    Assert::notNull($paymentMethod->getGatewayConfig());

    return $paymentMethod->getGatewayConfig()->getFactoryName() === 'offline';
  }
);
```

Zales0123: Good idea :

src/Action/Admin/RefundUnitsAction.php
```diff
@@ -46,6 +47,7 @@ public function __invoke(Request $request): Response
 
             $this->session->getFlashBag()->add('success', 'sylius_refund.units_successfully_refunded');
         } catch (CommandDispatchException $exception) {
+            Assert::notNull($exception->getPrevious());
```
bartoszpietrzak1994: Using Assert::notNull() may result in throwing unhandled InvalidArgumentException - what about simple and more user friendly solution:

```
if (null === $exception->getPrevious()) {
    $this->session->getFlashBag()->add('error', 'sylius.ui. unexpected_error_occurred');
}
```

wdyt?

Zales0123: The thing is, there will always be a _previous_ exception, as it's done automatically and silencing it could make more harm than good 😄 I'd leave it as it is (without it PHPStan is not passing)

PR #89, opened 2018-10-05
Switch to Symfony Messenger

Zales0123:
As Prooph service bus [has been deprecated](https://www.sasaprolic.com/2018/08/the-future-of-prooph-components.html) it's a good time to switch to a new, shiny [Symfony Messenger Component](https://symfony.com/doc/current/components/messenger.html).


src/Resources/config/app/config.yml
```diff
@@ -20,11 +20,14 @@ sylius_mailer:
...
-prooph_service_bus:
-    command_buses:
-        sylius_refund_command_bus: ~
-    event_buses:
-        sylius_refund_event_bus: ~
+framework:
+    messenger:
+        default_bus: messenger.bus.sylius_refund_command_bus
+        buses:
+            messenger.bus.sylius_refund_command_bus: ~
```
pamil: What about `sylius_refund_plugin.command_bus`?

tests/Behat/Context/Application/RefundingContext.php
```diff
@@ -104,7 +104,7 @@ public function decideToRefundPartFromProductWithPayment(
                 $paymentMethod->getId(),
                 ''
             ));
-        } catch (CommandDispatchException $exception) {
+        } catch (\Exception $exception) {
```
pamil: This looks very, very bad.

tests/Behat/Context/Application/RefundingContext.php
```diff
@@ -258,7 +265,7 @@ public function shouldStillBeAbleToRefundOrderShipment(PaymentMethodInterface $p
             $this->commandBus->dispatch(new RefundUnits(
                 $this->order->getNumber(), [], [new ShipmentRefund($shippingAdjustment->getId(), $remainingTotal)], $paymentMethod->getId(), ''
             ));
-        } catch (CommandDispatchException $exception) {
+        } catch (\Exception $exception) {
```
pamil: Can we typehint for a more specific exception? So that we can be sure that our application failed, not the infrastructure or anything else?

tests/Behat/Resources/services.xml
```diff
@@ -27,7 +27,7 @@
             <argument type="service" id="__symfony__.sylius.repository.order" />
             <argument type="service" id="__symfony__.sylius_refund.repository.refund" />
             <argument type="service" id="__symfony__.Sylius\RefundPlugin\Provider\RemainingTotalProvider" />
-            <argument type="service" id="__symfony__.prooph_service_bus.sylius_refund_command_bus" />
+            <argument type="service" id="__symfony__.message_bus" />
```
pamil: Shouldn't it use the specific command bus / event bus?

src/Resources/config/services/command_bus.xml
```diff
@@ -4,22 +4,20 @@
...
-
         <service id="Sylius\RefundPlugin\CommandHandler\RefundUnitsHandler">
             <argument type="service" id="Sylius\RefundPlugin\Refunder\OrderItemUnitsRefunder" />
             <argument type="service" id="Sylius\RefundPlugin\Refunder\OrderShipmentsRefunder" />
-            <argument type="service" id="prooph_service_bus.sylius_refund_event_bus" />
+            <argument type="service" id="messenger.bus.sylius_refund_event_bus" />
             <argument type="service" id="sylius.repository.order" />
             <argument type="service" id="Sylius\RefundPlugin\Validator\RefundUnitsCommandValidator" />
-            <tag name="prooph_service_bus.sylius_refund_command_bus.route_target" message-detection="true" />
+            <tag name="messenger.message_handler" />
```
pamil: On which message bus this handler is registered? Is it on both? First one?

src/Command/GenerateCreditMemo.php
```diff
@@ -4,47 +4,59 @@
...
+    private $shipments;
+
+    /** @var string */
+    private $comment;
+
+    public function __construct(string $orderNumber, int $total, array $units, array $shipments, string $comment)
     {
-        $this->init();
-        $this->setPayload([
-            'order_number' => $orderNumber,
```
pamil: This is 🔥

tests/Behat/Resources/services.xml
Zales0123: We can't, as only the default message bus (in fact it's alias `message_bus`) is public :|

tests/Behat/Context/Application/RefundingContext.php
Zales0123: Ok, I've figured out something :)

composer.json
```diff
@@ -22,8 +22,8 @@
         "php": "^7.2",
 
         "sylius/sylius": "^1.3",
-        "prooph/service-bus-symfony-bundle": "^0.7.0",
-        "knplabs/knp-snappy-bundle": "^1.5"
+        "knplabs/knp-snappy-bundle": "^1.5",
+        "symfony/messenger": "^4.1"
```
pamil: You also need symfony/framework-bundle ^4.1 to allow for `framework.messenger` configuration.

src/Resources/config/app/config.yml
```diff
@@ -20,11 +20,14 @@ sylius_mailer:
...
             template: "@SyliusRefundPlugin/Email/unitsRefunded.html.twig"
 
-prooph_service_bus:
-    command_buses:
-        sylius_refund_command_bus: ~
-    event_buses:
-        sylius_refund_event_bus: ~
+framework:
+    messenger:
+        default_bus: sylius_refund_plugin.command_bus
```
pamil: I copy-pasted it into similar PR in InvoicingPlugin, but I think we shouldn't set the default bus for the whole application here.

composer.json
Zales0123: Right 👍

src/Resources/config/app/config.yml
Zales0123: Totally agreed 👍

Zales0123: I've defined the default bus only [for the test env](https://github.com/Sylius/RefundPlugin/pull/89/files#diff-3cb4d392b9fb5e2d3cf5e66c4b081c7fR3). After all, we need this configuration only to be able to inject command bus to Behat contexts. This way it will be possible and also it won't be configured for the whole application 🐃

pamil:

PR #95, opened 2018-10-28
#78 Improve refund payment UI

### LINKED ISSUE #78: Improve refund payments UI
zrzut ekranu 2018-10-01 o 16 15 19

- [ ] ugly empty column if no action is possible
- [ ] status should be displayed as some label rather than regular string

laurent35240:
Fix #78 

* Do not remove cell if there is no action
* Display refund payment states with proper labels


laurent35240:
Here is a screenshot with the changes

order 000000020 sylius

Zales0123:
Great, thank you Laurent! 🥇

Issue #16, opened 2018-07-19
Sylius "refund" functionality


Zales0123:
We're working in this plugin for a proper refunding functionality. However, **Sylius** itself provide something like refunds - a simple state machine transitions with no logic at all.

It's quite confusing, as even though we provide "Refunds" button on order page, there is still a "Refund" button near the payment

<img width="284" alt="zrzut ekranu 2018-07-19 o 14 43 14" src="https://user-images.githubusercontent.com/6212718/42943219-705ba03e-8b62-11e8-8ea5-f5913aca0d6a.png">

and therefore payment can be marked as **refunded**

<img width="289" alt="zrzut ekranu 2018-07-19 o 14 43 22" src="https://user-images.githubusercontent.com/6212718/42943240-7f4b1c1e-8b62-11e8-8ed6-7dcbf8de207b.png">

The question is, should we somehow hide this behavior in our plugin? Or should it be fixed in **Sylius**?

cc @pjedrzejewski @CoderMaggie especially, @pamil @GSadee @bartoszpietrzak1994 of course as well

bartoszpietrzak1994:
It's for sure pretty misleading since the currently existing Refund button position is both quite obvious and intuitive. What's more, while Refund functionality placed in Sylius is related to payment, the feature that is currently implemented in RefundsPlugin is connected rather with the order itself. 

Considering these facts I would vote for Refund button removal. However, I'm afraid that removing it in the RefundPlugin's code would look like an ugly hack. On the other hand, I bet that there are plenty of shops based on Sylius that use current Refund feature. Hence, removing the button in Sylius/Sylius code doesn't seem like a perfect solution either.

What about putting both payment-related and order-related refunds in appropriate Refunds section to make everything clear?

CoderMaggie:
I would suggest a PR to Sylius/Sylius with this feature removal indicating in this PR that we have this plugin.

GSadee:
In first step, I would be for hiding payment refund button in this plugin. 
I'm not sure if we should remove this feature in Sylius/Sylius until the RefundPlugin functionalities are not in the Core.

alcaeus:
As for the "refund" button, just refunding a payment makes sense when you don't track anything else (which is the case for Sylius by default), but not when you do more with a refund. In my opinion, the button should either be hidden by the plugin or changed to lead to the actual refund functionality where it integrates nicely with the logic introduced by the bundle.

[...]

PR #96, opened 2018-10-31
Hide unneeded "refund" button

Zales0123:
Before:

zrzut ekranu 2018-10-31 o 11 00 42

After:

zrzut ekranu 2018-10-31 o 10 58 24

As we provide a sophisticated and quite expanded refunds system in this plugin, there is no need to keep this dummy **Refund** plugin anymore :) See https://github.com/Sylius/RefundPlugin/issues/16 for more details


pamil:
Thank you, @Zales0123! 🎉

Issue #98, opened 2018-11-05
Make it clearer that wkhtmltopdf is required


laurent35240:
While writing BDD tests locally, I got stuck because wkhtmlpdf was not installed on my local environment.
I think it will be a good idea to add into the README or somewhere else in documentation this requirement.

I think we can also remove the bin here https://github.com/Sylius/RefundPlugin/blob/master/tests/Application/etc/wkhtmltopdf and add instead installation of this bin in travis build.

Like that every one can more easily launch behat tests even if they are not on linux.

Let me know what you think about it and if it is ok for you I will write a PR for it.

lempere:
For newcomers, you can paste the wkhtmltopdf.exe in Application/etc (for windows environments) and past the tests.

bartoszpietrzak1994:
Hello laurent35240,

Indeed, the requirement is missing in the README.md file and it should be fixed as soon as possible.

When it comes to downloading the binary, we wanted to avoid build failing once the download website is unavailable. What's more, as far as I remember, after a few builds, Travis CI was blocked by the host for requesting the binary too often.

PR #104, opened 2018-11-25
#98 Add wkhtmltopdf requirement in README

laurent35240:


README.md
```diff
@@ -27,6 +27,10 @@ From Administrator's point of view, every Refund request results in creating two
 * Credit Memo - a document representing a list of refunded items (downloadable and sent to Customer via .pdf file)
 * Refund Payment - entity representing payment in favour of the Customer
 
+## Pre - requirements
+
+In order to be able to generate PDF files of refunds, [wkhtmltopdf](https://wkhtmltopdf.org/) must be installed.
```
Zales0123: Maybe we can mention that it's used by `KnpSnappyBundle`, not by plugin directly? And therefore it needs to be configured as mentioned in this bundle installation instruction https://github.com/KnpLabs/KnpSnappyBundle#configuration

PR #105, opened 2018-12-17
Fix build

Zales0123:


spec/CommandHandler/GenerateCreditMemoHandlerSpec.php
```diff
@@ -41,12 +41,8 @@ function it_generates_credit_memo(
...
         $creditMemoManager->flush()->shouldBeCalled();
 
-        $eventBus->dispatch(Argument::that(function (CreditMemoGenerated $event): bool {
-            return
-                $event->number() === '2018/01/000001' &&
-                $event->orderNumber() === '000666'
-            ;
-        }))->shouldBeCalled();
+        $event = new CreditMemoGenerated('2018/01/000001', '000666');
+        $eventBus->dispatch($event)->willReturn(new Envelope($event));
```
GSadee: Now, if I understand correctly, you are mocking the result instead of checking it as it was before, is it an intended change?
Maybe it should look like:
`$eventBus->dispatch($event)->shouldBeCalled();` or 
`$eventBus->dispatch($event)->shouldBeLike(new Envelope($event));` or sth similar?

PR #112, opened 2019-02-07
Fix CreditMemo migration

Zales0123:
There were some problems, it was incomplete and wrongly configured... so I've created it one more time from scratch, it should work fine :)


GSadee:

PR #113, opened 2019-02-07
Use migrations in Travis

Zales0123:
Based on https://github.com/Sylius/RefundPlugin/pull/112

Even though we've been providing migrations with this plugin for a long time, we didn't use them in a Travis CI configuration. And it was a mistake - it resulted in having them out-dated from time to time... and we could not be 100% sure they work 💃 

I've copied migration files from `Sylius-Standard` to make it possible to set up a database with `doctrine:migrations:migrate` rather than `doctrine:schema:create`. And of course it kind of unnecessary code in a `tests/Application`, but on the other hand, this test application should simulate a **Sylius-Standard** app which is using a `RefundPlugin`. And, as we know, each **Sylius-Standard** app is provided with a bunch of migrations inside 🐃 🏇


GSadee:
Thanks, Mateusz! :tada:

PR #115, opened 2019-02-08
Update to Sylius 1.4

Zales0123:


composer.json
```diff
@@ -21,7 +21,7 @@
     "require": {
         "php": "^7.2",
         "knplabs/knp-snappy-bundle": "^1.5",
-        "sylius/sylius": "1.3.*",
+        "sylius/sylius": "~1.4.0",
```
bartoszpietrzak1994: In other plugins we require `sylius/sylius` in `^1.4` version. Requiring it in `~1.4.0` version makes me wonder what approach is more suitable for us, WDYT?

Zales0123: `^1.4` 👌

behat.yml.dist
```diff
@@ -1,24 +1,47 @@
...
-                path: tests/Application/Kernel.php
-                bootstrap: ~
 
-        Lakion\Behat\MinkDebugExtension:
-            directory: etc/build
-            clean_start: false
-            screenshot: true
+        FriendsOfBehat\VariadicExtension: ~
+
+        FriendsOfBehat\SuiteSettingsExtension:
```
pamil: This can be removed - they're the default settings.

bartoszpietrzak1994:

PR #119, opened 2019-02-18
Better tests coverage

Zales0123:
Although I know that phpspec code coverage is a little bit controversial (in BDD you don't test, you describe, if you had not done it before, there is no point to go back), I think on Sylius plugins we can try to introduce some even more strict approach to testing than on Sylius itself. The variable there is a project size, we have pretty quick Travis build on plugins, much quicker than on Sylius, so if we slow down it a little bit with some code coverage check (it does not fail the build <yet?>, but displays a message) it won't be a big harm and it can give us potential benefits.

So, test coverage before this change:
zrzut ekranu 2019-02-18 o 14 27 07

After:
zrzut ekranu 2019-02-18 o 16 23 30

We missed few of them :)

I've also extracted a PHPSpec (with code coverage) build for a separated build. It makes sense IMO and can be used in the future for mutation tests (that's my dream 😍). 

I really believe that with such changes this plugin (together with other Sylius plugins) can became a real state-of-the-art technical wonder 🗽

### DIFF
--- a/phpspec.yml.dist
+++ b/phpspec.yml.dist
@@ -6,16 +6,16 @@ suites:
 extensions:
     LeanPHP\PhpSpec\CodeCoverage\CodeCoverageExtension:
         blacklist:
-            - src/Action
-            - src/DependencyInjection
-            - src/Entity
-            - src/File/TemporaryFileManager.php
-            - src/Form
-            - src/Menu
-            - src/StateResolver/OrderTransitions.php
-            - src/StateResolver/RefundPaymentTransitions.php
-            - src/SyliusRefundPlugin.php
-            - src/Twig
+            - src/Action #controllers should not be unit tests, as they only translates request to response (with some operations in the meantime)
+            - src/DependencyInjection #configuration
+            - src/Entity #anemic classes with no logic
+            - src/File/TemporaryFileManager.php #covered with phpunit
+            - src/Form #builders, no value in testing them
+            - src/Menu #builders, no value in testing them
+            - src/StateResolver/OrderTransitions.php #only constants
+            - src/StateResolver/RefundPaymentTransitions.php #only constants
+            - src/SyliusRefundPlugin.php #configuration
+            - src/Twig #covered with behat (strictly UI-oriented classes)
         format:
             - text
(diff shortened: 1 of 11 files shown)


phpspec.yml.dist
```diff
@@ -2,3 +2,21 @@ suites:
     sylius_refund:
         namespace: Sylius\RefundPlugin
         psr4_prefix: Sylius\RefundPlugin
+
+extensions:
+    LeanPHP\PhpSpec\CodeCoverage\CodeCoverageExtension:
+        blacklist:
```
pamil: I don't like the blacklist thing here, I would rather see what classes are not covered by PHPSpec rather than redefining them here. Could we also get PHPUnit coverage and merge them together? It seems a bit problematic for contributors, how would I know whether I should put some of my files here when I add a new feature?

Zales0123: I was thinking about such an approach, but I believe there are some few strong pros for a blacklist.

This list contains a list of catalogs we **for sure** don't want to cover by specs and we will not do it in any future (I think we agree it's unnecessary for controllers or dependency injection configuration). I see this blacklist as a configuration perfectly fit to specify such catalogs. Moreover, if we just allow them to be included in a code coverage report, it would be much harder to notice if any future PR does not contain the required specs (e.g. 100%->98.9% is more visible than 91.95%->90.85%). 

I agree that it can be not clear for contributors to determine "does my new class require PHPSpec test or not". Therefore, I think it should be clearly said in a proper README section when we should use a PHPSpec. In my opinion, classes that should be designed with PHPSpec are all services that contain any **logic**, any describable **behavior** - none of the blacklisted classes contain one, they're just configuration or raw data. We can also advice in a documentation "if you're not sure about writing a spec - write it". All mistakes (unneeded spec file, or missing test in an excluded directory) should be easily catchable during the code review process 👨‍💻 

Regarding PHPUnit - I don't know, I'll check it 😄 I'm not sure how it will behave, as we have only one `TemporaryFileManagerTest` file. But if it works with no problem, I'll add it :)

bartoszpietrzak1994:
Thanks, Mateusz! :tada:

Issue #111, opened 2019-01-29
[proposal] Use payment state machine on order instead of main state machine for refunded state.


prigal:
Actually this plugin set the "refunded" state on an order.

Refunded is something related to payment. An order can still be completed or whatever while it's payment is refunded.

Moreover Sylius already have refunded and partially refunded states on payment_state in order. 

Why not using them and keep order state unmodified ?

In our use case we will have "canceled" orders with refunded state on payment_state when order is canceled.

Zales0123:
It would be indeed nice to operate somehow on a payment state machine, as we have a relation between payment and refund since https://github.com/Sylius/RefundPlugin/pull/85. Needs to be considered before a stable release 🖖

prigal:
Imo we should not touch the sylius_payment.state state machine as long as payment can be completed on it's side and there is a refund_payment in front of it. 

But we should change order.payment_state state machine as long as it's an overview of all payments states in an order.

prigal:
Just for reference, if you want to work on it we did this in our fork https://github.com/printoclock/RefundPlugin. (as explained in an other ticket #110 , no time to PR properly but if it's not done by someone else in next future we will do it in a couple of months)

PR #123, opened 2019-02-22
Order payment state usage + "Partially refunded" state

Zales0123:
Fixes https://github.com/Sylius/RefundPlugin/issues/111

It was a mistake, to use the main order state machine to mark orders as refunded. Now "refunded" state is shown as an order payment state. Additionally, I've implemented marking order as "partially refunded" when *something* has been refunded.

zrzut ekranu 2019-02-21 o 16 38 49
zrzut ekranu 2019-02-22 o 11 38 27

### DIFF
--- a/features/having_order_partially_refunded.feature
+++ b/features/having_order_partially_refunded.feature
@@ -1,0 +1,22 @@
+@refunds
+Feature: Having order partially refunded
+    In order to note that part of the order total is refunded
+    As an Administrator
+    I want to have order partially refunded
+
+    Background:
+        Given the store operates on a single green channel in "United States"
+        And the store has a product "Mr. Meeseeks T-Shirt" priced at "$10.00"
+        And the store has "Galaxy Post" shipping method with "$10.00" fee
+        And the store allows paying with "Space money"
+        And there is a customer "rick.sanchez@wubba-lubba-dub-dub.com" that placed an order "#00000022"
+        And the customer bought 3 "Mr. Meeseeks T-Shirt" products
+        And the customer chose "Galaxy Post" shipping method to "United States" with "Space money" payment
+        And I am logged in as an administrator
+        And the order "#00000022" is already paid
+
+    @ui
+    Scenario: Having order partially refunded when some items are refunded
+        Given 1st "Mr. Meeseeks T-Shirt" product from order "#00000022" has already been refunded with "Space money" payment
+        When I browse orders
+        Then the order "#00000022" should have order payment state "Partially refunded"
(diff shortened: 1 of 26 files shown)


features/having_order_partially_refunded.feature
```diff
@@ -0,0 +1,28 @@
...
+    Scenario: Having order partially refunded when some items are refunded
+        Given 1st "Mr. Meeseeks T-Shirt" product from order "#00000022" has already been refunded with "Space money" payment
+        When I browse orders
+        Then the order "#00000022" should have order payment state "Partially refunded"
+
+    @ui
+    Scenario: Having order partially refunded when its shipping is refunded
+        Given shipment from order "#00000022" has already "$1.00" refunded with "Space money" payment
+        When I browse orders
+        Then the order "#00000022" should have order payment state "Partially refunded"
```
pamil: It looks like this `When` step is smuggling view logic inside business scenario - browsing orders do not change the state at all. I know we use a similar structure in Sylius, but we can do better in plugins 🎉 Doesn't need to be solved in this PR though.
```gherkin
When the shipment from order XXX is refunded $1 with "Space money" payment
Then the order XXX should have...
```

features/having_order_partially_refunded.feature
```diff
@@ -0,0 +1,28 @@
+@refunds
+Feature: Having order partially refunded
+    In order to note that part of the order total is refunded
+    As an Administrator
+    I want to have order partially refunded
```
bartoszpietrzak1994: It implies that the Administrator is the one that placed the order, IMO. What about `I want to be aware of the order payment state being partially refunded`?

src/Checker/OrderRefundingAvailabilityChecker.php
```diff
@@ -25,6 +25,9 @@ public function __invoke(string $orderNumber): bool
         $order = $this->orderRepository->findOneByNumber($orderNumber);
         Assert::notNull($order);
 
-        return $order->getPaymentState() === OrderPaymentStates::STATE_PAID;
+        return in_array(
+            $order->getPaymentState(),
+            [OrderPaymentStates::STATE_PAID, OrderPaymentStates::STATE_REFUNDED, OrderPaymentStates::STATE_PARTIALLY_REFUNDED]
```
bartoszpietrzak1994: Does it mean that refunding is still available if the order payment is in state `Refunded`?

Zales0123: That's because I wanted to make it possible to access the refund page when the order is fully refunded... but it won't be possible to refund anything because order's refunded total would be the same as order's total 😄 Probably the best choice would be to use a different service in a detection of possibility to go to the refund page and different for checking a refund action... On it.

bartoszpietrzak1994:

Issue #108, opened 2019-01-22
Sylius\RefundPlugin\Entity\Refund::getType() returns wrong type


maikrosenthal:
Sylius\RefundPlugin\Entity\Refund::getType() returns string instead of Sylius\RefundPlugin\Model\RefundType

PR #124, opened 2019-02-25
[HotFix] Return proper type in Refund type

Zales0123:
Fixes https://github.com/Sylius/RefundPlugin/issues/108

It's a hotfix, but not changing a current functionality - `Refund::$type` was indeed set in a database as a `string`. In fact, we should consider what to do with this property, as it's used nowhere 😄 and that was a reason why this bug has not been detected by tests. We should either remove this property or use it somewhere to avoid having a dead code ☠️

### DIFF
--- a/src/Entity/Refund.php
+++ b/src/Entity/Refund.php
@@ -21,7 +21,7 @@ class Refund implements RefundInterface
     /** @var int */
     private $refundedUnitId;
 
-    /** @var RefundType */
+    /** @var string */
     private $type;
 
     public function __construct(string $orderNumber, int $amount, int $refundedUnitId, RefundType $type)
@@ -29,7 +29,7 @@ class Refund implements RefundInterface
         $this->orderNumber = $orderNumber;
         $this->amount = $amount;
         $this->refundedUnitId = $refundedUnitId;
-        $this->type = $type;
+        $this->type = $type->__toString();
     }
 
     public function getId(): ?int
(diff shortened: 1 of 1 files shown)


pamil:
If it's not used, I'd remove it since it's not a stable package yet. In any other case, why don't we go with custom database type - https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/cookbook/custom-mapping-types.html?

Zales0123:
I didn't want to introduce a custom database type, because I wanted it to be a hotfix 😄 But after second consideration, I'm totally 👍 for removing it, we can always bring it back if it will be needed 🚀

Zales0123:
Ok, it seems it's not true, it's used 😄 https://github.com/Sylius/RefundPlugin/blob/e8bfb780c597899df450fdf67edbbb7103d18e36/src/Provider/RemainingTotalProvider.php#L38

@pamil What would you say for merging it, as it definitely fixes a problem that can be problematic... And providing a proper change in a separate PR? I can also create an issue for that and add it to the roadmap 🐃

bartoszpietrzak1994:
Since it's a hotfix I'd accept this solution. However, the ability to pass any string variable using 
the constructor of `Refund` class as a refund type makes the `RefundType` class even more useless - we have to reconsider this a bit :)

Zales0123:
Not any string, in constructor it's still required to pass `RefundType`, which is protected :)
https://github.com/Sylius/RefundPlugin/blob/master/src/Model/RefundType.php#L19

PR #130, opened 2019-04-17
Moved all entities to mapped-superclasses for better extensibility

kortwotze:
Can you also please tag a version of this for Sylius 1.3?

Closes #132.


kortwotze:
We need to finish a fix for the InvoicePlugin and port that over to this repository as well.

tests/Behat/Page/Admin/OrderRefundsPage.php
```diff
@@ -121,11 +121,7 @@ public function isPaymentMethodSelected(string $paymentMethodName): bool
             return false;
         }
 
-        if ($optionField->getText() != $paymentMethodName) {
```
lchrusciel: Could be part of separate PR so that we would merge it already :)

kortwotze: Yeah, that commit was actually just sitting in my local fork and got added to this PR. I will create another PR for that only. :)

kortwotze: For some reason this was already commited. Not sure why github thinks its still a change.

https://github.com/Sylius/RefundPlugin/commit/1d2d0bf9c7fdadc94afddc6783d649d0e5086312#diff-bccc45a8cb97fcb3c44b347a0a795920

Zales0123:
The main problem with this change is that we have a few [embeddables](https://github.com/Sylius/RefundPlugin/blob/8459e05904aa2822be05458f3bebd44959e3d62e/src/Resources/config/doctrine/CreditMemo.orm.xml#L24) used in the doctrine mapping... This is the same bug as here: https://github.com/Sylius/InvoicingPlugin/pull/97. Do you have any idea how it can be fixed? Every help would be helpful, as it for sure needs to be resolved.

kortwotze:
I currently have no idea how to fix that and I will probably not get the time to look into that in depth before our setup crashes. ;)

And yes, its the same as https://github.com/Sylius/InvoicingPlugin/pull/97 - I also mentioned this. Can we maybe tag these issues over Refund/Invoice/Rbac Plugin - because Rbac will probably get the same issues as soon as its starting to use embeddables.

Voting to keep the Invioice-Issue as the main discussion though :)

pamil:
One solution to rule them all: https://github.com/Sylius/SyliusResourceBundle/pull/89

pamil:
Merging it as it improves the current state of this plugin, though I'll investigate the issue with `CreditMemoSequence`.

pamil:
Thank you, Peter! :tada:

PR #131, opened 2019-04-24
Check events dispatching for real

Zales0123:
Without `shouldBeCalled()`, we could remove the `dispatch($event)` call and test were still passing 😢


bartoszpietrzak1994:

PR #132, opened 2019-04-30
MappedSuperclass with embeddables

Zales0123:
### Problem

It was detected on both [this](https://github.com/Sylius/RefundPlugin/pull/130) and [some other](https://github.com/Sylius/InvoicingPlugin/pull/97) plugin. If entities mapping are changed to `mapped-superclass`, embeddable relations are not resolved at all (columns are not added to entities' tables). This is, of course, a pity, as embeddables are highly useful for resolving some problems.

My first thought was that our `ORMMappedSuperClassSubscriber` from *ResourceBundle* is doing something wrong, but I've realized that class metadata passed to its `loadClassMetadata` function already has embeddable columns set (or **not set** in our case).

### Idea

The problem is in [this line from doctrine/orm package](https://github.com/doctrine/orm/blob/2.6/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php#L165). Apparently, mapped superclasses are not meant to be used with embeddables for some reason :/ Our subscriber is converting mapped superclasses to entities (with `convertToEntityIfNeeded` function) but it listens to `loadClassMetadata` which is thrown **after** embeddables resolving 🚀 At the end, the best idea is to use `convertToEntityIfNeeded` function before processing embeddables, but after [loading metadata for class](https://github.com/doctrine/orm/blob/2.6/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php#L151). 

### Solution

I've decorated the `MappingDriverChain` class to throw an event just after metadata is loaded for a class (which is still before resolving embeddables) and overwritten the `ORMMappedSuperClassSubscriber` to listen to this event and do the same as for `loadClassMetadata` event. It works 🎉 but has some drawbacks:

- doesn't it break anything? Only because tests are passing, I can't be 100% sure that the logic is correct
- our custom mapping driver *must* be named `DriverChain` not to break any gedmo extension (which is ridiculous, see https://github.com/Atlantic18/DoctrineExtensions/blob/0b7bdbefd3d166def27928dcd62ab67c11c8f172/lib/Gedmo/Mapping/ExtensionMetadataFactory.php#L136)
- because of the custom driver, we need to implement some functions that are not essential for resolving this problem (like `getDrivers()` for example)

The separated question is, if this fix is correct, where should it be done? I believe we should fix it in Sylius, but I open a PR here to show that this [is working](https://travis-ci.org/Sylius/RefundPlugin/jobs/526392238#L985) and to not litter the main Sylius repo with some potentially invalid solution 😄 

I need your help and opinion very much, to see if it's in any part the way we want to go to fix this frustrating problem 🖖 

cc @pamil @lchrusciel @bartoszpietrzak1994 @kortwotze

### DIFF
--- a/src/Doctrine/Mapping/Driver/DriverChain.php
+++ b/src/Doctrine/Mapping/Driver/DriverChain.php
@@ -1,0 +1,59 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sylius\RefundPlugin\Doctrine\Mapping\Driver;
+
+use Doctrine\Common\Persistence\Mapping\ClassMetadata;
+use Doctrine\Common\Persistence\Mapping\Driver\MappingDriver;
+use Doctrine\Common\Persistence\Mapping\Driver\MappingDriverChain as BaseMappingDriverChain;
+use Doctrine\ORM\EntityManager;
+use Doctrine\ORM\Event\LoadClassMetadataEventArgs;
+
+class DriverChain implements MappingDriver
+{
+    /** @var BaseMappingDriverChain */
+    private $baseMappingDriverChain;
+
+    /** @var EntityManager */
+    private $entityManager;
+
+    public function __construct(BaseMappingDriverChain $baseMappingDriverChain, EntityManager $entityManager)
+    {
+        $this->baseMappingDriverChain = $baseMappingDriverChain;
+        $this->entityManager = $entityManager;
+    }
+
+    public function loadMetadataForClass($className, ClassMetadata $metadata): void
+    {
+        $this->baseMappingDriverChain->loadMetadataForClass($className, $metadata);
+
+        if (
+            $this->entityManager->getEventManager()->hasListeners('preEmbeddableResolve') &&
+            !empty($metadata->embeddedClasses)
+        ) {
+            $eventArgs = new LoadClassMetadataEventArgs($metadata, $this->entityManager);
+            $this->entityManager->getEventManager()->dispatchEvent('preEmbeddableResolve', $eventArgs);
+        }
+    }
+
+    public function getAllClassNames()
+    {
+        return $this->baseMappingDriverChain->getAllClassNames();
+    }
+
+    public function isTransient($className)
+    {
+        return $this->baseMappingDriverChain->isTransient($className);
+    }
+
+    public function getDrivers(): array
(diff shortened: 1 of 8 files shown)


src/Doctrine/Mapping/Driver/DriverChain.php
```diff
@@ -0,0 +1,59 @@
...
+            $this->entityManager->getEventManager()->dispatchEvent('preEmbeddableResolve', $eventArgs);
+        }
+    }
+
+    public function getAllClassNames()
+    {
+        return $this->baseMappingDriverChain->getAllClassNames();
+    }
+
+    public function isTransient($className)
```
GSadee: You could add return type

src/Doctrine/Mapping/Driver/DriverChain.php
```diff
@@ -0,0 +1,59 @@
...
+
+namespace Sylius\RefundPlugin\Doctrine\Mapping\Driver;
+
+use Doctrine\Common\Persistence\Mapping\ClassMetadata;
+use Doctrine\Common\Persistence\Mapping\Driver\MappingDriver;
+use Doctrine\Common\Persistence\Mapping\Driver\MappingDriverChain as BaseMappingDriverChain;
+use Doctrine\ORM\EntityManager;
+use Doctrine\ORM\Event\LoadClassMetadataEventArgs;
+
+class DriverChain implements MappingDriver
```
lchrusciel: Can we finalize it?

kortwotze:
I will try this PR as soon as we have finalized the 1.4.x upgrade to our systems which can possibly be done this week. We will see. :) Thanks for this! :heart: 

To be honest, I think this is not resolving any broken things, but rather adding a lot of convenience for developers to be able to extend standard sylius entities with proper means and not overriding all mappings for a plugin.

In addition I think the solution to convert mapped-superclasses to entities before embeddables are resolved should be correct. I see no usecase right now where someone would do shenanigans in the same place of execution. :)

pamil:
Took this code and transformed it into https://github.com/Sylius/SyliusResourceBundle/pull/89.

Zales0123:
Closing, as it should work after https://github.com/Sylius/SyliusResourceBundle/pull/89 in #130

PR #133, opened 2019-05-06
Symfony-extension version temporarily locked

bartoszpietrzak1994:
This should be an applicable fix until the newest versions of https://github.com/FriendsOfBehat/SymfonyExtension/issues/80 are fixed.


Zales0123:
Thanks, Bartosz! :tada:

PR #143, opened 2019-06-05
Remove conflicts and update to Messenger 4.3

pamil:


.travis.yml
```diff
@@ -81,7 +81,7 @@ jobs:
                 - composer analyse
 
                 - vendor/bin/phpunit
-                - vendor/bin/behat --strict -vvv --no-interaction --tags="~@todo" || vendor/bin/behat --strict -vvv --no-interaction --rerun --tags="~@todo"
+                - vendor/bin/behat --strict -vvv --no-interaction --tags="~@todo && @application"
```
lchrusciel: Do we want to run application tests only?

pamil: Good catch, I used it to speed up feedback loop while working on it, gonna revert it now.

PR #146, opened 2019-07-11
[Admin] Unify order link in Credit Memos

Tomanhez:
**Before:**
Screenshot 2019-07-11 at 11 43 17

**After:**
Screenshot 2019-07-12 at 07 45 08

**After click in link:**
Screenshot 2019-07-11 at 13 56 14
Screenshot 2019-07-11 at 13 56 23

### DIFF
--- a/src/Resources/views/Order/Grid/Field/number.html.twig
+++ b/src/Resources/views/Order/Grid/Field/number.html.twig
@@ -1,0 +1,1 @@
+<a href="{{ path('sylius_refund_admin_order_show_by_number', {'number': data}) }}">#{{ data }}</a>
--- a/src/Resources/config/admin_routing.yml
+++ b/src/Resources/config/admin_routing.yml
@@ -74,3 +74,13 @@ sylius_refund_admin_credit_memo_send:
     methods: [GET]
     defaults:
         _controller: Sylius\RefundPlugin\Action\Admin\SendCreditMemoAction
+
+sylius_refund_admin_order_show_by_number:
+    path: /orders/by-number/{number}
+    methods: [GET]
+    defaults:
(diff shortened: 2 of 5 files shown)


src/Resources/views/Order/Grid/Field/number.html.twig
```diff
@@ -0,0 +1 @@
+<a href="{{ path('sylius_admin_order_show', {'id' : sylius_refund_get_order_id_by_order_number(data.orderNumber) })|raw }}">#{{ data.orderNumber }}</a>
```
pamil: I'm not really convinced to the idea of having a template function `sylius_refund_get_order_id_by_order_number` access the repository every time it has to show order number link just to generate the link itself.

We have an endpoint that shows order page for given order ID, what if we could create an endpoint that takes an order number, gets the order ID and redirects you to the former? This way we do not need any additional SQL queries for every order link (20-100 per page). What do you think? This has to be done in Sylius/Sylius though, but I think we're in process of making shortcuts like this there also.

pamil: Actually, we can temporarily create that action in this plugin, if there's no need for it in Sylius itself.

Tomanhez: You are a right ! but it must be in Sylius/Sylius. In this moment I think that I haven't better way

Tomanhez: I changed the twig function on redirect because twig function when render link in every row send query to database, now db is used when clicking in link. Thanks for advice

GSadee:
Change the column name from `Order number` to `Order` to be consistent

GSadee:

PR #156, opened 2019-10-08
Fix refund buttons alignment

kulczy:
Zrzut ekranu 2019-10-8 o 14 04 44


GSadee:
Thank you, Szymon! :1st_place_medal:

PR #157, opened 2019-10-14
Require Sylius 1.5

JakobTolkemit:
This plugin requires `sylius/resource-bundle: ^1.6.0-RC.2` as well as `sylius/sylius: ^1.4`.
But Sylius 1.4 requires `sylius/resource-bundle: ~1.4.4`, which means this plugin is not actually compatible with Sylius 1.4.


pamil:
Thank you, Jakob! :1st_place_medal:

PR #158, opened 2019-10-14
In order refund page disable buttons if all order is refunded

Tomanhez:
Screenshot 2019-10-15 at 10 07 23

### DIFF
--- a/src/Resources/views/orderRefunds.html.twig
+++ b/src/Resources/views/orderRefunds.html.twig
@@ -6,6 +6,10 @@
 
 {% set customer = order.customer %}
 
+{% set disableButton = '' %}
+
+{% set disableButton = order.paymentState == 'refunded' ? 'disabled' : '' %}
+
 {% block content %}
     {% include '@SyliusRefundPlugin/_header.html.twig' %}
     {% include '@SyliusAdmin/Order/Show/_breadcrumb.html.twig' %}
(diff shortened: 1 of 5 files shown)


src/Resources/views/orderRefunds.html.twig
```diff
@@ -6,6 +6,12 @@
 
 {% set customer = order.customer %}
 
+{% set disableButton = '' %}
+
+{% if order.paymentState == 'refunded' %}
+    {% set disableButton = 'disabled' %}
+{% endif %}
```
pamil: Can be refactored into:

```twig
{% set disableButton = order.paymentState == 'refunded' ? 'disabled' : '' %}
```

src/Resources/views/orderRefunds.html.twig
```diff
@@ -6,6 +6,10 @@
 
 {% set customer = order.customer %}
 
+{% set disableButton = '' %}
```
pamil: This should be not needed anymore.

tests/Behat/Context/Ui/RefundingContext.php
```diff
@@ -283,6 +283,16 @@ public function shouldStillBeAbleToRefundOrderShipment(): void
         Assert::true($this->orderRefundsPage->isOrderShipmentAvailableToRefund());
     }
 
+    /**
+     * @Then I should not be able to refund anything
+     */
+    public function iShouldNotBeAbleToRefundAnything(): void
+    {
+        Assert::false($this->orderRefundsPage->canRefund());
```
lchrusciel: Do we need all three methods here?

src/Resources/views/orderRefunds.html.twig
```diff
@@ -6,6 +6,8 @@
 
 {% set customer = order.customer %}
```
GSadee: Unnecessary blank line

src/Resources/views/orderRefunds.html.twig
```diff
@@ -6,6 +6,8 @@
...
 
+{% set disableButton = order.paymentState == 'refunded' ? 'disabled' : '' %}
```
GSadee: Use constant state instead of string `refunded`

GSadee:

PR #163, opened 2019-10-21
Make entity properties protected

JakobTolkemit:
Extending entities is a pain if the properties are private. Doctrine complains about them not existing on the entity.


Zales0123:
Thanks, Jakob! :tada:

JakobTolkemit:
When will this be released? We kinda need it.

PR #164, opened 2019-11-15
Set default messenger bus

Jeroen-G:
Prevents the following error: `Invalid configuration for path "framework.messenger": You must specify the "default_bus" if you define more than one bus.`

### DIFF
--- a/src/Resources/config/app/config.yml
+++ b/src/Resources/config/app/config.yml
@@ -19,6 +19,7 @@ sylius_mailer:
 
 framework:
     messenger:
+        default_bus: sylius_refund_plugin.command_bus
         buses:
             sylius_refund_plugin.command_bus: ~
             sylius_refund_plugin.event_bus:


Zales0123:
Hello Jeroen! 

Specifying the default message but is one of the steps of [plugin installation](https://github.com/Sylius/RefundPlugin#installation). As far as I remember, there is a problem if you use two plugins that have the messenger configuration inside and both of them define a default bus... Could you check it out, will there be some problems in such a situation? Thanks in advance! 🖖

loevgaard:
The default bus should be specified in the app, yes

Jeroen-G:
Ah I see, I did not see that installation guide because the plugin got installed as part of Sylius Plus.

Zales0123:
So this is definitely the issue for the SyliusPlus installation instructions. I've raised it up in the team :) Cheers 🚀

PR #167, opened 2019-11-29
[Refund] Do not expose exception messages to the customer

lchrusciel:


GSadee:
Thank you, Łukasz! :1st_place_medal:

PR #171, opened 2020-01-08
[WIP] Credit memo order relation

AdamKasp:
Add a relation between order and credit memo as a first step to have easily navigation between refund/credit memo and order. 
replaced by #172

### DIFF
--- a/src/Resources/config/admin_routing.yml
+++ b/src/Resources/config/admin_routing.yml
@@ -25,15 +25,15 @@ sylius_refund_credit_memo:
     type: sylius.resource
 
 sylius_refund_order_credit_memos_list:
-    path: /orders/{orderNumber}/credit-memos
+    path: /orders/{orderId}/credit-memos
     methods: [GET]
     defaults:
         _controller: sylius_refund.controller.credit_memo:indexAction
         _sylius:
             template: "@SyliusRefundPlugin/Order/Admin/CreditMemo/list.html.twig"
             repository:
-                method: findByOrderNumber
-                arguments: $orderNumber
+                method: findByOrder
+                arguments: $orderId
 
 sylius_refund_credit_memo_details:
     path: /orders/{orderNumber}/credit-memos/{id}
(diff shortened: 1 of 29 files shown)


src/Resources/config/admin_routing.yml
```diff
@@ -44,7 +44,6 @@ sylius_refund_credit_memo_details:
             template: "@SyliusRefundPlugin/Order/Admin/CreditMemo/details.html.twig"
             criteria:
                 id: $id
-                orderNumber: $orderNumber
```
lchrusciel: It should stay I suppose

src/Resources/config/shop_routing.yml
```diff
@@ -1,13 +1,13 @@
 sylius_refund_plugin_shop_order_credit_memos_partial:
-    path: /_partial/credit-memos/{orderNumber}/
+    path: /_partial/credit-memos/{orderId}/
```
lchrusciel: I'm not sure if 
```suggestion
    path: /_partial/orders/{id}/credit-memos/
```
wouldn't be more obvious. /cc @Zales0123  @GSadee 

This path suggests IMHO only one credit memo.

tests/Application/config/packages/doctrine.yaml
```diff
@@ -12,3 +12,6 @@ doctrine:
         charset: UTF8
 
         url: '%env(resolve:DATABASE_URL)%'
+    orm:
+        resolve_target_entities:
+            Sylius\Component\Core\Model\OrderInterface: Sylius\Component\Core\Model\Order
```
lchrusciel: Why did you add it here?

Zales0123: For some reason, it was not working without... I was doing a quick PoC so I've added it and decided to resolve it later. @AdamKasp now is "later" 😄

src/Resources/config/shop_routing.yml
Zales0123: 👍

PR #175, opened 2020-01-15
Configuring supported gateways instead of hardcoding them

Zales0123:
It would much easier to customize supported payment methods. Thanks @loevgaard for the idea!


lchrusciel:

PR #179, opened 2020-01-23
[CreditMemo] Rework credit memo

GSadee:
image

Zrzut ekranu 2020-01-25 o 20 20 49


features/seeing_details_of_credit_memo.feature
```diff
@@ -0,0 +1,59 @@
...
+        Then it should have sequential number generated from current date
+        And it should be issued in "United States" channel
+        And its subtotal should be "5.00"
+        And it should contain 1 "PHP T-Shirt" product with "5.00" net value, "0.50" tax amount and "5.50" gross value in "USD" currency
+        And it should have a tax item "US VAT (10%)" with amount "0.50" in "USD" currency
+        And its total should be "5.50" in "USD" currency
+
+    @ui @application
+    Scenario: Seeing the details of generated credit memo with partial shipment price
+        Given shipment from order "#00000023" has already "$4.50" refunded with "Space money" payment
```
CoderMaggie: 😱 whaaat

```suggestion
        Given the "#00000023" order's shipping cost already has a refund of "$4.50" with "Space money" payment
```

Is this payment specification needed in this step? Looks like unnecessary fuss 🤷‍♀

features/seeing_details_of_credit_memo.feature
```diff
@@ -0,0 +1,59 @@
...
+        And it should contain 1 "PHP T-Shirt" product with "10.00" net value, "1.00" tax amount and "11.00" gross value in "USD" currency
+        And it should contain 1 "Symfony Mug" product with "17.50" net value, "3.50" tax amount and "21.00" gross value in "USD" currency
+        And its subtotal should be "27.50" in "USD" currency
+        And it should have a tax item "US VAT (10%)" with amount "1.00" in "USD" currency
+        And it should have a tax item "VAT (20%)" with amount "3.50" in "USD" currency
+        And its total should be "32.00" in "USD" currency
+
+    @ui @application
+    Scenario: Seeing the details of generated credit memo with partial price
+        Given 1st "PHP T-Shirt" product from order "#00000022" has already "$5.00" refunded with "Space money" payment
```
CoderMaggie:
```suggestion
        Given the 1st "PHP T-Shirt" product from order "#00000022" already has a refund of"$5.00" with "Space money" payment
```

features/seeing_details_of_credit_memo.feature
```diff
@@ -0,0 +1,59 @@
...
+        When I browse the details of the only credit memo generated for order "#00000022"
+        Then it should have sequential number generated from current date
+        And it should be issued in "United States" channel
+        And its subtotal should be "5.00"
+        And it should contain 1 "PHP T-Shirt" product with "5.00" net value, "0.50" tax amount and "5.50" gross value in "USD" currency
+        And it should have a tax item "US VAT (10%)" with amount "0.50" in "USD" currency
+        And its total should be "5.50" in "USD" currency
```
CoderMaggie: Why does this step differ from [this one](https://github.com/Sylius/RefundPlugin/pull/179/files#diff-bfa80ae7010f420ce914820afb19dbf8R47) regarding currency?

features/seeing_details_of_credit_memo_with_promotion_applied.feature
```diff
@@ -0,0 +1,37 @@
+@refunds
+Feature: Seeing the details of generated credit memo for order with the promotion applied
```
CoderMaggie:
```suggestion
Feature: Viewing details of a credit memo for order with a promotion applied

File: features/seeing_details_of_credit_memo_with_promotion_applied.feature [hunk 5]
```diff
@@ -0,0 +1,37 @@
...
+    In order to be aware of what order units have been refunded
+    As an Administrator
+    I want to be able to view details of generated credit memo
```
CoderMaggie:
```suggestion
    I want to be able to view details of a credit memo
```

features/seeing_details_of_credit_memo_with_promotion_applied.feature
```diff
@@ -0,0 +1,37 @@
...
+        And this promotion gives "$3.99" discount to every order with quantity at least 4
+        And there is a customer "rick.sanchez@wubba-lubba-dub-dub.com" that placed an order "#00000022"
+        And the customer bought 4 "PHP T-Shirt" products
+        And the customer "Rick Sanchez" addressed it to "Seaside Fwy", "90802" "Los Angeles" in the "United States" with identical billing address
+        And the customer chose "Galaxy Post" shipping method with "Space money" payment
+        And the order "#00000022" is already paid
+        And I am logged in as an administrator
+
+    @ui @application
+    Scenario: Seeing the details of generated credit memo for order with the promotion applied
```
CoderMaggie:
```suggestion
    Scenario: Viewing details of a credit memo issued for an order with a promotion applied
```

features/seeing_details_of_credit_memo.feature
GSadee: I don't understand the question, subtotal is a net value, but total gross value, probably I should change the naming

CoderMaggie: And its subtotal should be "5.00" 
And its total should be "5.50" **in "USD" currency** 

imo it should be either:
```
And its subtotal should be "5.00" 
And its total should be "5.50"
```

or 

```
And its subtotal should be "5.00" in "USD" currency
And its total should be "5.50" in "USD" currency
```

it looks inconsistent now.

UPGRADE.md
```diff
@@ -1,7 +1,19 @@
+### UPGRADE FROM 1.0.0-RC.1 TO 1.0.0-RC.2
+
+1. `CreditMemoUnit` changed to `LineItem` which is a resource entity now
+
+2. `CreditMemoUnitGeneratorInterface` changed to `LineItemsConverterInterface` and moved from `Generator` to `Converter` directory
```
pamil: What about using FQCN here so that directory changes become namespace changes?

migrations/Version20200125182414.php
```diff
@@ -0,0 +1,32 @@
...
+
+        $this->addSql('CREATE TABLE sylius_refund_credit_memo_line_items (credit_memo_id VARCHAR(255) NOT NULL, line_item_id INT NOT NULL, INDEX IDX_1453B90E8E574316 (credit_memo_id), UNIQUE INDEX UNIQ_1453B90EA7CBD339 (line_item_id), PRIMARY KEY(credit_memo_id, line_item_id)) DEFAULT CHARACTER SET UTF8 COLLATE `UTF8_unicode_ci` ENGINE = InnoDB');
+        $this->addSql('CREATE TABLE sylius_refund_line_item (id INT AUTO_INCREMENT NOT NULL, name VARCHAR(255) NOT NULL, quantity INT NOT NULL, unit_net_price INT NOT NULL, unit_gross_price INT NOT NULL, net_value INT NOT NULL, gross_value INT NOT NULL, tax_amount INT NOT NULL, tax_rate VARCHAR(255) DEFAULT NULL, PRIMARY KEY(id)) DEFAULT CHARACTER SET UTF8 COLLATE `UTF8_unicode_ci` ENGINE = InnoDB');
+        $this->addSql('ALTER TABLE sylius_refund_credit_memo_line_items ADD CONSTRAINT FK_1453B90E8E574316 FOREIGN KEY (credit_memo_id) REFERENCES sylius_refund_credit_memo (id)');
+        $this->addSql('ALTER TABLE sylius_refund_credit_memo_line_items ADD CONSTRAINT FK_1453B90EA7CBD339 FOREIGN KEY (line_item_id) REFERENCES sylius_refund_line_item (id)');
+        $this->addSql('ALTER TABLE sylius_refund_credit_memo DROP units');
```
pamil: Could we migrate those units to line items?

src/Converter/LineItemsConverterInterface.php
```diff
@@ -0,0 +1,19 @@
...
+use Sylius\RefundPlugin\Model\UnitRefundInterface;
+
+interface LineItemsConverterInterface
+{
+    /**
+     * @param array<UnitRefundInterface> $units
+     *
+     * @return Collection|LineItemInterface[]
+     */
+    public function convert(array $units): Collection;
```
pamil: Is there any reason why we need to return a Doctrine's Collection? Could it be just a plain old PHP array?

src/Converter/LineItemsConverter.php
```diff
@@ -0,0 +1,107 @@
...
+     */
+    private function getTaxRate(Collection $taxAdjustments): ?string
+    {
+        if ($taxAdjustments->isEmpty()) {
+            return null;
+        }
+
+        $label = $taxAdjustments->first()->getLabel();
+
+        if (preg_match('#\((.*?)\)#', $label, $matches)) {
```
lchrusciel: o.O What has happened here?

I don't have an idea what we try to achieve here.

src/Converter/LineItemsConverterInterface.php
```diff
@@ -0,0 +1,19 @@
...
+namespace Sylius\RefundPlugin\Converter;
+
+use Doctrine\Common\Collections\Collection;
+use Sylius\RefundPlugin\Entity\LineItemInterface;
+use Sylius\RefundPlugin\Model\UnitRefundInterface;
+
+interface LineItemsConverterInterface
+{
+    /**
+     * @param array<UnitRefundInterface> $units
```
pamil:
```suggestion
     * @param UnitRefundInterface[] $units
```

src/Converter/LineItemsConverter.php
pamil: Using more common `/` around the regexp and adding a named matching group:

```
if (preg_match('/\((?P<rate>.*?)\)/', $label, $matches)) {
    return $matches['rate'];
}
```

src/Converter/LineItemsConverter.php
```diff
@@ -0,0 +1,107 @@
...
+        $this->orderItemUnitRepository = $orderItemUnitRepository;
+    }
+
+    public function convert(array $units): Collection
+    {
+        $lineItems = new ArrayCollection();
+
+        /** @var UnitRefundInterface $unitRefund */
+        foreach ($units as $unitRefund) {
+            /** @var OrderItemUnitInterface $orderItemUnit */
```
pamil: `find` can return null as well, we can be sure it's `OrderItemUnitInterface` only after asserting it's not null.

```suggestion
            /** @var OrderItemUnitInterface|null $orderItemUnit */
```

src/Generator/CreditMemoGenerator.php
```diff
@@ -37,15 +38,15 @@ final class CreditMemoGenerator implements CreditMemoGeneratorInterface
     private $uuidCreditMemoIdentifierGenerator;
 
     public function __construct(
-        CreditMemoUnitGeneratorInterface $orderItemUnitCreditMemoUnitGenerator,
-        CreditMemoUnitGeneratorInterface $shipmentCreditMemoUnitGenerator,
+        LineItemsConverterInterface $lineItemsConverter,
+        LineItemsConverterInterface $shipmentLineItemsConverter,
```
lchrusciel: As the order of these services is important, I would recommend creating two interfaces instead of one. Otherwise, errors may be really misleading (when `units` would be passed to `shipmentLineItemsConvertera` and `shipments` to `lineItemsConverter`)

src/Converter/LineItemsConverter.php
```diff
@@ -0,0 +1,107 @@
...
+     */
+    private function addLineItem(LineItemInterface $newLineItem, Collection $lineItems): Collection
+    {
+        /** @var LineItemInterface $lineItem */
+        foreach ($lineItems as $lineItem) {
+            if (
+                $lineItem->name() === $newLineItem->name() &&
+                $lineItem->unitNetPrice() === $newLineItem->unitNetPrice() &&
+                $lineItem->unitGrossPrice() === $newLineItem->unitGrossPrice() &&
+                $lineItem->taxRate() === $newLineItem->taxRate()
```
pamil: If we have a method to merge line items, why don't we add one to compare them?

src/Converter/ShipmentLineItemsConverter.php
```diff
@@ -0,0 +1,52 @@
...
+        $this->adjustmentRepository = $adjustmentRepository;
+    }
+
+    public function convert(array $units): Collection
+    {
+        $lineItems = new ArrayCollection();
+
+        /** @var UnitRefundInterface $unitRefund */
+        foreach ($units as $unitRefund) {
+            /** @var AdjustmentInterface $shippingAdjustment */
```
pamil:
```suggestion
            /** @var AdjustmentInterface|null $shippingAdjustment */
```

src/Converter/ShipmentLineItemsConverter.php
```diff
@@ -0,0 +1,52 @@
...
+            Assert::lessThanEq($unitRefund->total(), $shippingAdjustment->getAmount());
+
+            $lineItems->add(new LineItem(
+                $shippingAdjustment->getLabel(),
+                1,
+                $unitRefund->total(),
+                $unitRefund->total(),
+                $unitRefund->total(),
+                $unitRefund->total(),
+                0
```
pamil: What about taxes applied on the shipment?

src/Entity/CreditMemo.php
```diff
@@ -114,14 +115,9 @@ public function getChannel(): ChannelInterface
...
-    public function getUnits(): array
+    public function getLineItems(): Collection
     {
-        $units = [];
-        foreach ($this->units as $unit) {
-            $units[] = CreditMemoUnit::unserialize($unit);
-        }
-
-        return $units;
+        return $this->lineItems;
```
pamil: Feels good 👍

src/Entity/LineItem.php
```diff
@@ -0,0 +1,114 @@
...
+    {
+        return $this->taxAmount;
+    }
+
+    public function taxRate(): ?string
+    {
+        return $this->taxRate;
+    }
+
+    public function merge(LineItemInterface $newLineItem): void
```
pamil: I'd add a sanity check to make sure we are not merging two completely different line items.

src/Resources/config/doctrine/CreditMemo.orm.xml
```diff
@@ -11,33 +11,46 @@
...
         </many-to-one>
 
         <many-to-one field="order" target-entity="Sylius\Component\Order\Model\OrderInterface">
-            <join-column name="order_id" referenced-column-name="id" nullable="true" />
+            <join-column name="order_id" />
         </many-to-one>
 
+        <many-to-many field="lineItems" target-entity="Sylius\RefundPlugin\Entity\LineItemInterface">
+            <cascade>
+                <cascade-persist />
```
lchrusciel: I'm wondering if we shouldn't cascade all operations

src/Resources/views/Download/creditMemo.html.twig
```diff
@@ -56,59 +66,69 @@
...
-            <td>{{ unit.productName }}</td>
-            <td>{{ unit.taxesTotal|sylius_format_money(creditMemo.currencyCode, creditMemo.localeCode) }}</td>
-            <td>{{ unit.total|sylius_format_money(creditMemo.currencyCode, creditMemo.localeCode) }}</td>
-        </tr>
+        {% for item in creditMemo.lineItems %}
+            <tr class="item">
+                <td>{{ loop.index }}</td>
+                <td>{{ item.name }}</td>
+                <td>{{ item.quantity }}</td>
+                <td>{{ (item.unitNetPrice/100)|number_format(2) }}</td>
```
lchrusciel: Why didn't you used `sylius_format_money`?

src/Generator/TaxItemsGeneratorInterface.php
```diff
@@ -4,15 +4,16 @@
...
 interface TaxItemsGeneratorInterface
 {
     /**
-     * @param array<UnitRefundInterface> $units
+     * @param Collection|LineItemInterface[] $lineItems
      *
      * @return array<TaxItemInterface>
      */
-    public function generate(array $units): array;
+    public function generate(Collection $lineItems): array;
```
pamil: Is there any reason to accept a `Collection` instead of an array?

tests/Behat/Context/Transform/PriceContext.php
```diff
@@ -0,0 +1,34 @@
...
+            $price *= -1;
+        }
+
+        return $price;
+    }
+
+    /** @throws \InvalidArgumentException */
+    private function validatePriceString(string $price): void
+    {
+        if (!(bool) preg_match('/^\d+(?:\.\d{1,2})?$/', $price)) {
```
lchrusciel: Can't we use this regular expression in `@Transform`?

UPGRADE.md
```diff
@@ -1,7 +1,19 @@
+### UPGRADE FROM 1.0.0-RC.1 TO 1.0.0-RC.2
+
+1. `CreditMemoUnit` changed to `LineItem` which is a resource entity now
```
lchrusciel:
```suggestion
1. `CreditMemoUnit` was changed to `LineItem` which is a resource entity now
```

src/Resources/views/Download/creditMemo.html.twig
pamil: I'm not sure whether it'll render prices like `4.20` correctly - `number_format` might cut it to `4.2` if not for those parentheses.

```suggestion
                <td>{{ '%0.2f'|format(item.unitNetPrice/100) }}</td>
```

This will make sure there are always two digits to the right of the decimal point. I'm not too convinced to any option though.

tests/Behat/Context/Transform/PriceContext.php
```diff
@@ -0,0 +1,34 @@
...
+declare(strict_types=1);
+
+namespace Tests\Sylius\RefundPlugin\Behat\Context\Transform;
+
+use Behat\Behat\Context\Context;
+
+final class PriceContext implements Context
+{
+    /**
+     * @Transform /^"(\-)?((?:\d+\.)?\d+)"$/
```
pamil: What's the reason behind matching quotation marks, not just the value inside them?

features/refunding_part_of_order_unit.feature
```diff
@@ -28,23 +28,23 @@ Feature: Refunding a part of an order unit
 
     @ui @application
     Scenario: Refunding the whole order unit price after partial refund
-        Given 1st "Mr. Meeseeks T-Shirt" product from order "#00000022" has already "$5.00" refunded with "Space money" payment
+        Given the 1st "Mr. Meeseeks T-Shirt" product from order "#00000022" already has a refund of "$5.00" with "Space money" payment
```
pamil:
```suggestion
        Given the 1st "Mr. Meeseeks T-Shirt" product from order "#00000022" has been already refunded "$5.00" with "Space money" payment
```

pamil:
```suggestion
        Given the 1st "Mr. Meeseeks T-Shirt" product from order "#00000022" has a refund of "$5.00" with "Space money" payment
```

features/having_credit_memo_generated.feature
```diff
@@ -7,62 +7,19 @@ Feature: Having credit memo generated
...
-        And the store has a product "Mr. Meeseeks T-Shirt" priced at "$10.00"
-        And it belongs to "Clothes" tax category
+        And the store has a product "PHP T-Shirt" priced at "$10.00"
         And the store has "Galaxy Post" shipping method with "$10.00" fee
         And the store allows paying with "Space money"
-        And there is a promotion "Anatomy Park Promotion"
-        And this promotion gives "$1.00" off on every product with minimum price at "$5.00"
         And there is a customer "rick.sanchez@wubba-lubba-dub-dub.com" that placed an order "#00000022"
-        And the customer bought 2 "Mr. Meeseeks T-Shirt" products
+        And the customer bought 2 "PHP T-Shirt" products
```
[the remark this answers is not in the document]
GSadee: That was just simpler 🎉

src/Resources/views/Download/creditMemo.html.twig
GSadee: Because I want to have prices without currency 😢

src/Converter/LineItemsConverter.php
GSadee: Yup, because it is temporary cancer ♋️ to get tax rate from tax adjustment label, I was wondering if it is ok for now

src/Generator/CreditMemoGenerator.php
GSadee: I'm not convinced to that, the input and output are the same and having the same interface is logical to me

tests/Behat/Context/Transform/PriceContext.php
GSadee: To be honest, I'm not sure, but I will check that

migrations/Version20200125182414.php
GSadee: IMO no, because I do not have enough information and also the old credit memos shouldn't be changed

src/Converter/LineItemsConverterInterface.php
GSadee: There is no specific reason, a line item is just an entity now, so using a collection here seems to be appropriate imo

src/Converter/ShipmentLineItemsConverter.php
GSadee: It's a good question, I haven't changed the current behaviour for that in this PR

src/Resources/views/Download/creditMemo.html.twig
Zales0123: So the other question - why we want here prices without currency? 😄

src/Generator/TaxItemsGeneratorInterface.php
GSadee: The same as with https://github.com/Sylius/RefundPlugin/pull/179#discussion_r372227908

tests/Behat/Context/Transform/PriceContext.php
GSadee: I didn't change the implementation from `LexicalContext` from Sylius, but I guess that it had been written that way to this transformer doesn't catch some unwanted strings (as order number in scenarios from this plugin)

src/Resources/views/Download/creditMemo.html.twig
GSadee: Because credit memo should look like this 😃 /cc @CoderMaggie

src/Converter/ShipmentLineItemsConverter.php
CoderMaggie: I've split that into a separate task

spec/Entity/CreditMemoSpec.php
```diff
@@ -37,11 +38,16 @@ function let(ChannelInterface $channel, OrderInterface $order): void
...
 
-    function it_implements_credit_memo_interface(): void
+    function it_implements_a_credit_memo_interface(): void
     {
         $this->shouldImplement(CreditMemoInterface::class);
     }
 
+    function it_implements_a_resource_interface(): void
+    {
+        $this->shouldImplement(ResourceInterface::class);
```
Zales0123: Do we need to check it? `CreditMemoInterface` extends `ResourceInterface`... I'm just not sure does it give us any value 🤔

src/Converter/LineItemsConverter.php
```diff
@@ -0,0 +1,102 @@
...
+            $orderItemUnit = $this->orderItemUnitRepository->find($unitRefund->id());
+            Assert::notNull($orderItemUnit);
+            Assert::lessThanEq($unitRefund->total(), $orderItemUnit->getTotal());
+
+            /** @var OrderItemInterface $orderItem */
+            $orderItem = $orderItemUnit->getOrderItem();
+
+            $grossValue = $unitRefund->total();
+            $taxAmount = (int) ($grossValue * $orderItemUnit->getTaxTotal() / $orderItemUnit->getTotal());
+            $netValue = $grossValue - $taxAmount;
```
Zales0123: I feel that there is a lot of things we could extract there - even if they're done only there, it still does not seem nice 😢 Maybe we could have some VO like `CreditMemoValues` which would calculate these variables? Or some `LineItemFactory`? It's not a must-have, **for now,** for me, but would definitely improve a DX 🖖

src/Converter/LineItemsConverterInterface.php
Zales0123: If it could be an array I would vote for this 👍 Collection seems a little bit coupled with the ORM and entities, this interface could be used anywhere else. We can still just use the result of this function as an `ArrayCollection` argument 💃

src/Entity/LineItem.php
```diff
@@ -0,0 +1,134 @@
...
+    /** @var int */
+    protected $netValue;
+
+    /** @var int */
+    protected $grossValue;
+
+    /** @var int */
+    protected $taxAmount;
+
+    /** @var ?string */
```
Zales0123: `string|null`?

src/Entity/LineItem.php
```diff
@@ -0,0 +1,134 @@
...
+        if (
+            $this->name() === $lineItem->name() &&
+            $this->unitNetPrice() === $lineItem->unitNetPrice() &&
+            $this->unitGrossPrice() === $lineItem->unitGrossPrice() &&
+            $this->taxRate() === $lineItem->taxRate()
+        ) {
+            return true;
+        }
+
+        return false;
```
Zales0123:
```php
return
    $this->name() === $lineItem->name() &&
    $this->unitNetPrice() === $lineItem->unitNetPrice() &&
    $this->unitGrossPrice() === $lineItem->unitGrossPrice() &&
    $this->taxRate() === $lineItem->taxRate()
;
```
wdyt?

src/Generator/CreditMemoGenerator.php
Zales0123: Maybe the best would be the queue with priorities accepting only array of specific units 💃 But I would leave it as it is right now

src/Converter/LineItemsConverter.php
```diff
@@ -0,0 +1,102 @@
...
+     */
+    private function getTaxRate(Collection $taxAdjustments): ?string
+    {
+        if ($taxAdjustments->isEmpty()) {
+            return null;
+        }
+
+        $label = $taxAdjustments->first()->getLabel();
+
+        if (preg_match('/\((.*?)\)/', $label, $matches)) {
```
Zales0123: As many people are not familiar with regexp (or even hate them 😄) maybe we could add some comment about what does it do (or extract it to the private method 💃)?

src/Converter/LineItemsConverter.php
GSadee: For now, I've extracted creating line item to the separate private method, but I agree that in future some refactor could improve a DX

src/Converter/LineItemsConverter.php
```diff
@@ -0,0 +1,106 @@
...
```
Zales0123: We should definitely think how to handle it better (either here on on `Sylius/Sylius`) but let's just open an issue for that :)

src/Resources/views/Download/creditMemo.html.twig
CoderMaggie: Confirmed by PO 👩‍⚖

migrations/Version20200125182414.php
pamil: They'll be changed after we drop this column though

pamil:
Thanks, Grzegorz! :tada:

PR #181, opened 2020-01-31
Extract tax rate provider from line items converter

GSadee:
After #179

### DIFF
--- a/src/Provider/TaxRateProvider.php
+++ b/src/Provider/TaxRateProvider.php
@@ -1,0 +1,30 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sylius\RefundPlugin\Provider;
+
+use Doctrine\Common\Collections\Collection;
+use Sylius\Component\Core\Model\AdjustmentInterface;
+use Sylius\Component\Core\Model\OrderItemUnitInterface;
+
+final class TaxRateProvider implements TaxRateProviderInterface
+{
+    public function provide(OrderItemUnitInterface $orderItemUnit): ?string
+    {
+        /** @var Collection|AdjustmentInterface[] $taxAdjustments */
+        $taxAdjustments = $orderItemUnit->getAdjustments(AdjustmentInterface::TAX_ADJUSTMENT);
+
+        if ($taxAdjustments->isEmpty()) {
+            return null;
+        }
+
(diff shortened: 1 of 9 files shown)


src/Provider/TaxRateProvider.php
```diff
@@ -0,0 +1,30 @@
...
+
+namespace Sylius\RefundPlugin\Provider;
+
+use Doctrine\Common\Collections\Collection;
+use Sylius\Component\Core\Model\AdjustmentInterface;
+use Sylius\Component\Core\Model\OrderItemUnitInterface;
+
+final class TaxRateProvider implements TaxRateProviderInterface
+{
+    public function provide(OrderItemUnitInterface $orderItemUnit): ?string
```
Zales0123: I wonder, should we pass there `$orderItemUnit` or maybe tax adjustment would be more suitable? The order item unit may or may not contain tax adjustment (therefore, we're forced to allow a null return). I feel like this check should be done before and not even call this service if we have no tax adjustment to process 🤔

src/Provider/TaxRateProvider.php
```diff
@@ -0,0 +1,30 @@
...
+
+declare(strict_types=1);
+
+namespace Sylius\RefundPlugin\Provider;
+
+use Doctrine\Common\Collections\Collection;
+use Sylius\Component\Core\Model\AdjustmentInterface;
+use Sylius\Component\Core\Model\OrderItemUnitInterface;
+
+final class TaxRateProvider implements TaxRateProviderInterface
```
Zales0123: Maybe `LabelBasedTaxRateProvider`?

src/Provider/TaxRateProvider.php
GSadee: I guess that the goal is to get a tax rate for a refunded order item unit. In the implementation of this service, there is an assumption that we take the first adjustment, but this may not be the best solution for all cases and for me, this kind of service should decide about that. So, passing here `$orderItemUnit` seems to me a better option to achieve the original goal.

Zales0123: Makes sense 👍

Zales0123:
Thank you, Grzegorz! :1st_place_medal:

PR #187, opened 2020-02-10
[Credit Memo] Fix shipping cost calculation

Tomanhez:


features/viewing_details_of_credit_memo.feature
```diff
@@ -55,3 +57,12 @@ Feature: Viewing details of a credit memo
...
         And its total should be "4.50" in "USD" currency
+
+    @todo
+    Scenario: Viewing details of a credit memo issued for a shipping cost refund
+        Given the "#00000022" order's shipping cost already has a refund of "$11.50" with "Space money" payment
+        When I browse the details of the only credit memo generated for order "#00000022"
+        Then it should have sequential number generated from current date
+        And it should contain 1 "Galaxy Post" shipment with "10.00" net value, "1.50" tax amount and "11.50" gross value in "USD" currency
+        And it should be issued in "United States" channel
+        And its total should be "11.50" in "USD" currency
```
GSadee: Add before this step:
```
And it should contain a tax item "15%" with amount "1.50" in "USD" currency
```

src/Converter/ShipmentLineItemsConverter.php
```diff
@@ -43,14 +51,21 @@ private function convertUnitRefundToLineItem(UnitRefundInterface $unitRefund): L
         Assert::notNull($shippingAdjustment);
         Assert::lessThanEq($unitRefund->total(), $shippingAdjustment->getAmount());
 
+        /** @var ShipmentInterface $shipment */
+        $shipment = $this->shipmentRepository->find(['id' => $unitRefund->id()]);
+        Assert::notNull($shipment);
+        /** @var AdjustmentInterface $adjustment */
+        $adjustment = $shipment->getUnits()->first()->getAdjustments()->first();
```
lchrusciel: Are you sure that the first adjustment of the first unit of a shipment always contains information about shipping costs?

grimzy:
While this PR is now outdated; the issue persists; it doesn't seems that we can refund the shipping tax (or rather the order's 'tax' adjustment); even when refunding every item in the order.

Additionally (consequently?), when refunding every item in the order, the order's status remains `partially_refunded`. Is this because the adjustments on the order are not refunded?

Am I using the refund plugin wrong? Is there a configuration I'm missing somewhere? How do we refund the taxes associated with an order?

GSadee:
Closing in favour of #264

Issue #188, opened 2020-02-14
Refund payment's state machine states should be in lower case


diimpp:
https://github.com/Sylius/RefundPlugin/blob/1886a1bb2894599c93955016f270610ac2bdfb5c/src/Resources/config/app/config.yml#L49

```
      New --> new
      Completed --> completed
```

PR #190, opened 2020-02-17
Fix to state label

tuala:
Quick fix to state label


Zales0123:
Hello @tuala! I believe this change is related to [this issue](https://github.com/Sylius/RefundPlugin/issues/188)? If yes, it should be definitely fixed on the state machine level (I mean - states in a state machine should be defined with lowercase).

GSadee:
I believe it has been already fixed in https://github.com/Sylius/RefundPlugin/pull/197

PR #198, opened 2020-03-05
Minor restyle of order credit memo list ui

diimpp:
- Updates color schema for credit memo `Show/Download` buttons to neutral gray/secondary action blue. Buttons are grouped. 
- Replaces custom channel widget (Visually incorrect) with `@SyliusAdmin/Common/_channel.html.twig' ` 
- Adds `issuedAt` value to header of credit memo details page (No issued at value without it at all)
- Cleans trailing whitespaces in README

Before
image

After
image

P.S. I've renamed `Show` button to `Details`, which looks better to me, but this change is optional and I can revert it.

### DIFF
--- a/src/Resources/views/Download/creditMemo.html.twig
+++ b/src/Resources/views/Download/creditMemo.html.twig
@@ -1,7 +1,7 @@
 {% set to = creditMemo.to %}
 {% set from = creditMemo.from %}
 
-<html lang="en">
+<html lang="{{ creditMemo.localeCode }}">
 <head>
     <meta charset="utf-8">
     <style>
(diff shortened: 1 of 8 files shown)


src/Resources/views/Download/creditMemo.html.twig
```diff
@@ -31,9 +31,8 @@
                         <td colspan="2" class="title">
                             <img src="https://demo.sylius.com/assets/shop/img/logo.png" style="width:300px;">
                         </td>
-
                         <td>
-                            {{ 'sylius_refund.ui.issued_at'|trans }}: {{ creditMemo.issuedAt|date('Y-m-d H:i:s') }}<br>
+                            {{ 'sylius_refund.ui.issued_at'|trans }}: {% if creditMemo.issuedAt is not null %}{{ creditMemo.issuedAt|date('Y-m-d H:i:s') }}{% endif %}<br>
```
diimpp: I've noticed, that `CreditMemo->getIssuedAt()` has not nullable return typehint, therefore this check is unnecessary here and in other places.

So doctrine mapping for CreditMemo issuedAt needs to updated to non-nullable and getIssuedAt method should clone the value or work with immutable datatime.

lchrusciel:
Thanks, Dmitri! :tada:

PR #199, opened 2020-03-06
Fix getId/id return typehints for entities

### LINKED ISSUE #189: Wrong getId type-hint, it should be nullable
Hello,

Those entities won't have an ID after construction as they those IDs are assigned by doctrine, but at the same time `getId`/`id` type-hints for them are too restrictive.

```patch
--- getId(): int
+++ getId(): ?int
```
https://github.com/Sylius/RefundPlugin/blob/master/src/Entity/CreditMemoSequence.php
https://github.com/Sylius/RefundPlugin/blob/master/src/Entity/CustomerBillingData.php
https://github.com/Sylius/RefundPlugin/blob/master/src/Entity/LineItem.php
https://github.com/Sylius/RefundPlugin/blob/master/src/Entity/RefundPayment.php
https://github.com/Sylius/RefundPlugin/blob/master/src/Entity/ShopBillingData.php
https://github.com/Sylius/RefundPlugin/blob/master/src/Entity/TaxItem.php

`LineItem` and `TaxItem` have both `getId` and `id` methods, but `CustomerBillingData` and `ShopBillingData` doesn't, while they have similar methods schema. 

I guess spec would need to be changed as well with something like
```php
    function it_has_no_id_by_default(): void
    {
        $this->getId()->shouldReturn(null);
    }
```
https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/Payment/spec/Model/PaymentSpec.php#L27

Other thing that I've noticed, that `LineItem` and `TaxItem` `id` doctrine definition doesn't have `AUTO` keyword, if it means anything.
https://github.com/Sylius/RefundPlugin/blob/master/src/Resources/config/doctrine/LineItem.orm.xml#L6

diimpp:
Fixes #189

### DIFF
--- a/spec/Entity/CreditMemoSequenceSpec.php
+++ b/spec/Entity/CreditMemoSequenceSpec.php
@@ -14,6 +14,11 @@ final class CreditMemoSequenceSpec extends ObjectBehavior
         $this->shouldImplement(SequenceInterface::class);
     }
 
+    function it_has_no_id_by_default(): void
+    {
+        $this->getId()->shouldReturn(null);
+    }
+
     function it_has_incrementable_index(): void
     {
         $this->getIndex()->shouldReturn(0);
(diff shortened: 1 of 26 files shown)


lchrusciel:

Issue #191, opened 2020-02-18
Credit Memo number generator requires refactor


diimpp:
Hi,

There are multiple problems with given number generator, functionally, styling and idiomatic wise.

- `NumberGenerator` interface's `generate` method doesn't have arguments, like `OrderNumberGeneratorInterface`'s `generate` have `$order`, which leads to problem by which data to generate the number.
- https://github.com/Sylius/RefundPlugin/blob/master/src/Generator/NumberGenerator.php `NumberGenerator` interface doesn't have `Interface` part in it's name
- Sylius traditional placement is `src/NumberGenerator`, name should be `CreditMemoNumberGeneratorInterface`
-  ~[SequentialNumberGenerator](https://github.com/Sylius/RefundPlugin/blob/master/src/Generator/SequentialNumberGenerator.php#L52) uses `DateTime('now')` as source of the date, but it should take date from CreditMemo IssuedAt. This is outright bug and was done so, because `NumberGenerator` interface's `generate` doesn't have CreditMemo as argument~ I've noticed how CreditMemo is instanced and why CreditMemo is not supplied to number generator, but issue still holds true: `generate(OrderInterface $order, \DateTimeInterface $issuedAt)`  Something like this maybe?
- ~[Entity\SequenceInterface::getId](https://github.com/Sylius/RefundPlugin/blob/master/src/Entity/SequenceInterface.php#L9) should be nullable~ **Done**
- ~[Entity\SequenceInterface](https://github.com/Sylius/RefundPlugin/blob/master/src/Entity/SequenceInterface.php) should be named as `CreditMemoSequenceInterface`, this plugin is not a place to offer generic sequence interface~ **Done**
- ~[Needs strict comparison](https://github.com/Sylius/RefundPlugin/blob/master/src/Generator/SequentialNumberGenerator.php#L77)~ **Done**
- [DI definition is not friendly for extension](https://github.com/Sylius/RefundPlugin/blob/master/src/Resources/config/services/generator.xml#L7). Interface should be used as base service name or regular non-fqcn string name.

Some of the entities of this plugin, including `CreditMemoSequence` don't implement `ResourceInterface`. Is it some kind of policy? I would have preferred them to be regular resources, whenever possible.

PR #200, opened 2020-03-06
Misc refactor

### LINKED ISSUE #192: Replace Address::getFirstName . getLastName with Address::getFullName
https://github.com/Sylius/RefundPlugin/blob/1886a1bb2894599c93955016f270610ac2bdfb5c/src/Generator/CreditMemoGenerator.php#L89

diimpp:
Tried to collect BC changes in single PR.

Fixes #192
References #191 

- Renames SequenceInterface to CreditMemoSequenceInterface (Like OrderSequenceInterface, InvoiceSequenceInterface)
- Makes CreditMemo `number` unique and `issuedAt` non nullable (In doctrine mappings, for the rest of the app it's already holds true)
- Rename `sylius_refund_payment` to `sylius_refund_refund_payment` (Table was named incorrectly)
- Splits CustomerBillingData `customerName` into `firstName` and `lastName` (As in Addressing\Model\Address, Customer\Model\Customer, Invoicing\Entity\BillingData)
- Change CreditMemo `issuedAt` to datetime_immutable (Note, that I've converted CurrentDateTimeInterface to CurrentDateTimeImmutableInterface due to phpstan unability to handle properly \DateTimeInterface in context of \DateTime and \DateTimeImmutable https://github.com/phpstan/phpstan/issues/314 https://phpstan.org/r/d15041c9-a230-4b18-a627-fc859ea00838 )


diimpp:
@lchrusciel Hi, sorry, it's still in work. I've moved PR from draft too early, because travis wasn't building properly, but it was the other issue.

diimpp:
@lchrusciel @Zales0123 it's ready for review. Let me know if you have any questions.

This PR is in fact 6 unrelated PRs, so please review each commit separately. Only difference is last two commits, because last commit is a supporting hack for phpstan.

P.S. There is another problematic code, which is `src/StateResolver` (No `can` checks, non-idiomatic files placement, `Order->state` state resolver doesn't work for me at all, but I didn't look too closely at it yet), but I won't be looking at it for the moment.

Cheers,
Dmitri.

diimpp:
@lchrusciel thank you for update. I do plan to change Refund/RefundPayment/CreditMemo mappings drastically, so it may be worth waiting for next PR anyway.

Current implementation with Refund/RefundPayment/CreditMemo, which are not related between each other is not usable for my company. We need `CreditMemo` to be generated only after `RefundPayment` is completed. Another case, we need to different between `Refund`'s as when they were done and how they were grouped and after the fact of the refund, not at the refund event itself. (Another point is just technically not feasible approach. Let's say event fails for whatever reasons, then Refunds are created, but CreditMemo/RefundPayment probably not. And there is no way to re-send this event.)

lchrusciel:

PR #202, opened 2020-03-12
[DB] Remove `COMMENT` database configuration

lchrusciel:


diimpp:
I would vote to drop this ENUM thing entirely. Isn't used in Sylius, so no reason to use over here.

diimpp:
@lchrusciel Would you agree to drop ENUM package? I can add PR for that later on.

lchrusciel:
I'm in favor of it, but I would like to gather the second vote here. /cc @Zales0123

Zales0123:
Dropping the enum -> simplification -> profit 👌

Zales0123:
Thanks, Łukasz! :tada:

lchrusciel:
@diimpp do you think, you could find time to work on this enum removal? I would like to publish a new version of RefundPlugin and this is one of the changes, that could be included

diimpp:
@lchrusciel yes, but I think state machine code for Order->state==refunded is broken (Maybe not true, not really tested by me) and there is matter of Refund/RefundPayment/CreditMemo design issue. It's on lower priority for me, but I will check it tomorrow and let you know if I can do it early.

lchrusciel:
Therefore, I will probably publish the current state as it is, as some useful refactorings were done already.

diimpp:
@lchrusciel, it look like a hefty task, many method signatures using this `RefundType` enum, 41 entry in the source code.

So I would be better to go ahead with my other refactor first and look at this afterwards. Some of the classes, that are using this enum may change drastically.

PR #203, opened 2020-04-06
Fix 'Unknown "locale" filter' error

### LINKED ISSUE #201: [Sylius 1.7] locale filter not found
While testing Sylius 1.7 I encountered an error into the administration on `sylius_refund_order_refunds_list` route

```
Unknown "locale" filter. Did you mean "locale_name", "sylius_locale_name", "sylius_locale_country"?
```

This occurred because of a filter used in `vendor/sylius/refund-plugin/src/Resources/views/orderRefunds.html.twig` line 29.

I don't know if it was an old Sylius Twig filter or a Twig add-on one, but maybe this filter can be deleted ?

dunglas:
Close #201. According to my tests, it's the only change needed to make this plugin compatible with 1.7

### DIFF
--- a/src/Resources/views/_header.html.twig
+++ b/src/Resources/views/_header.html.twig
@@ -18,7 +18,7 @@
                             {{ order.currencyCode }}
                         </div>
                         <div class="item">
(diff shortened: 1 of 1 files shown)


lchrusciel:
Thanks, Kévin! :1st_place_medal:

PR #204, opened 2020-04-07
Migrations: fix compatibility with MySQL 8

dunglas:
Without this change, the migrations fail on MySQL 8.


lchrusciel:
Thank you, Kévin! :tada:

PR #209, opened 2020-05-25
Fix wrong type error if the error in some cases

### LINKED ISSUE #208: RufundUnitsHandler fails on PHP error somewhere in the process
Those lines are a problem if at some point there's a PHP error:
https://github.com/Sylius/RefundPlugin/blob/85ea56038b537d84d46a9a9212ff13aba8eaaa6c/src/Action/Admin/RefundUnitsAction.php#L61-L63

It will fail without rendering the actual error.

Suggestion:

```diff
-private function provideErrorMessage(\Exception $previousException): void
+private function provideErrorMessage(\Throwable $previousException): void
```

Nek-:
Problem
-------

The raised error may be a throwable since the user is supposed to add its custom code with handlers. But since this is type hinted Exception, the real failure will be hidden by this new TypeError.

Solution
--------

Accept Throwable instead.

Fixes #208

### DIFF
--- a/src/Action/Admin/RefundUnitsAction.php
+++ b/src/Action/Admin/RefundUnitsAction.php
@@ -70,7 +70,7 @@ final class RefundUnitsAction
         ));
     }
 
-    private function provideErrorMessage(\Exception $previousException): void
+    private function provideErrorMessage(\Throwable $previousException): void
     {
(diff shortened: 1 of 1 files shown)


GSadee:
Thanks, Maxime! :1st_place_medal:

PR #205, opened 2020-04-24
Switch to Doctrine/Migrations 3.0

pamil:
Related to https://github.com/Sylius/Sylius/pull/11389.

PR #210, opened 2020-05-26
Update phpspec to ^6.0 version

Nek-:
Problem
-------

phpspec was set in version ^4. This version was not compatible with
PHP 7.4. So if you have the latest version of PHP on your computer you
cannot develop on this plugin.

And the upgrade was blocking because of another dependency.

Solution
--------

Update phpspec to ^6 version. Change phpspec-code-coverage to the
friends-of-phpspec version which allows us to use phpspec ^6.

Bonus
-----

Add php 7.4 on travis ci.


GSadee:
Closing as it has been done in https://github.com/Sylius/RefundPlugin/pull/205

PR #211, opened 2020-05-26
Make the CI green again

Nek-:
Contains #210 

Note: I removed support for PHP 7.2 because Sylius does not officially support PHP 7.2 in last releases and it makes it complicated to deal with compatibility between all version on the tested app. (basically, the configuration is different, so we should have a directory of conf by php version)

I think the current state is fine enough.

Possible improvement:
- Remove the migration that handle all sylius stuff and replace it by `doctrine:schema:update --force`, but this needs some effort to remove the refund bundle before.
- Refactor the whole thing to drop selenium :grimacing:


Nek-:
@jacquesbh you misread it, I didn't change any migration of the plugin, it's just a migration for the test application (those migrations should not even exist actually, they are here just because it's easier to deal with the plugin installation with them, but they are non-sense and make test non-reliable)

jacquesbh:
My bad! Sorry for the noise then :p

tests/Application/config/packages/security_checker.yaml
```diff
@@ -1,9 +1,9 @@
-services:
-    SensioLabs\Security\SecurityChecker:
-        public: false
-
-    SensioLabs\Security\Command\SecurityCheckerCommand:
-        arguments: ['@SensioLabs\Security\SecurityChecker']
-        public: false
-        tags:
-            - { name: console.command, command: 'security:check' }
+#services:
```
lchrusciel, line 1: Why did you comment this out instead of removing this file?

tests/Application/src/Migrations/Version1.php
```diff
@@ -4,185 +4,198 @@
...
 
-use Doctrine\Migrations\AbstractMigration;
 use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
 
 /**
  * Auto-generated Migration: Please modify to your needs!
  */
-class Version20170912085504 extends AbstractMigration
+final class Version1 extends AbstractMigration
```
lchrusciel, line 13: If this migration is used only in test env I'm ok with this change. Nonetheless, why can't we leave the previous implementation and add new migrations on top of them?

Also, we cannot move to `schema:upgrade --force` as test migrations are compound with main migrations and we need to run all of them to test if they work. 

Also, we are working with a new approach to migration: https://github.com/Sylius/Sylius/pull/11389 Hopefully it will reduce such problems in the future.

tests/Application/config/packages/security_checker.yaml
Nek-: I forgot to remove it nice catch.

tests/Application/src/Migrations/Version1.php
Nek-: The problem is that this migration contains all the sylius database changes, which will often change. Actually I'm in favor of a schema:update instead of using migrations for that.

jacquesbh: I agree that for me the best way to install a plugin is to run the `doctrine:migration:diff` command, for Migrations.

Nek-: After re-reading the thread I think we're all talking of different things.

First, @jacquesbh this migration is only for the behat test environment. It's not related to the migrations of the plugin. This is another debate.

Second, I talk about `do:schema:update` for this behat application only. I understand running the plugin migration is an actual test. And I also understand that since the plugin is running at all times the `do:schema:up` will include db changes of the plugin. But my point is that we could enable dynamically the plugin while building, and after the generation of the database.


Finally, I forgot to answer clearly to this message:

> Nonetheless, why can't we leave the previous implementation and add new migrations on top of them?

One of the issues is that the migration is not up to date inside the app. I can ofc just add a new one but I think it's far more clear to have only 1 migration. This will hopefully avoid @jacquesbh [to misunderstand](https://github.com/Sylius/RefundPlugin/pull/211#issuecomment-634695467) the thing for example ;).

lchrusciel: According to the one migration, we used to copy additional, required migration from the UPGRADE.md file. You can see it here: https://github.com/Sylius/InvoicingPlugin/commit/f75f7763709d3a5fd1acc77d2f5a65114d49a1a8. The reasoning behind it is that at the end tests application - is just Sylius application and we also test upgrade files this way. What is more, we are keeping them in sync, which reduce cost of maintenance.

Nek-:
Closing because the Sylius team does not want this approach. I let them fix issue the way they want. I saw on some other repositories they refactored entirely the tests and it's not the goal of this PR.

lchrusciel:
Hey @Nek-, of course, I totally respect your decision. I hope we will see each other on some other PR's.

However, I cannot guarantee we will pick up your work before 1.8. 

In the meantime, if someone would like to finish this PR, the main blocker was with the commented configuration. Preferably, I would recommend to bring back old migrations and just added new ones. The rest is not needed to merge it.

Issue #186, opened 2020-02-04
Factories not implementing FactoryInterface


loevgaard:
When I run `php bin/console lint:container` I get:

`Invalid definition for service "sylius_refund.controller.refund": argument 5 of "Sylius\Bundle\ResourceBundle\Controller\ResourceController::__construct" accepts "Sylius\Component\Resource\Factory\FactoryInterface", "Sylius\RefundPlugin\Factory\RefundFactory" passed.`

Possibly [these factories](https://github.com/Sylius/RefundPlugin/tree/master/src/Factory) should implement the Sylius factory interface?

PR #212, opened 2020-05-26
Remove wrong factory from configuration

Nek-:
Problem
-------

The RefundFactory is not a factory designed to be used in sylius resources. As a side effect it generates invalid services and may fail.

Solution
--------

Removing the factory configuration. Since it generates wrong service configuration, it's probably not used anywhere.

Fix https://github.com/Sylius/RefundPlugin/issues/186


mamazu:
Can be closed as it is replaced by #223

GSadee:
Closing in favour of #223

PR #214, opened 2020-05-26
[DX] Easier tests for RefundUnits message

Nek-:
Problem
-------

The `OrderItemUnitRefund` and `ShipmentRefund` implements interfaces but the `RefundUnits` does not use the given interface. Since those classes are final the assertion was not allowing mocking classes.

Solution
--------

Asserting on interfaces instead of final classes.

### DIFF
--- a/src/Command/RefundUnits.php
+++ b/src/Command/RefundUnits.php
@@ -15,6 +15,7 @@ namespace Sylius\RefundPlugin\Command;
 
 use Sylius\RefundPlugin\Model\OrderItemUnitRefund;
 use Sylius\RefundPlugin\Model\ShipmentRefund;
+use Sylius\RefundPlugin\Model\UnitRefundInterface;
 use Webmozart\Assert\Assert;
 
 class RefundUnits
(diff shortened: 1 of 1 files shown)


src/Command/RefundUnits.php
```diff
@@ -27,8 +28,8 @@ final class RefundUnits
 
     public function __construct(string $orderNumber, array $units, array $shipments, int $paymentMethodId, string $comment)
     {
-        Assert::allIsInstanceOf($units, OrderItemUnitRefund::class);
-        Assert::allIsInstanceOf($shipments, ShipmentRefund::class);
+        Assert::allIsInstanceOf($units, UnitRefundInterface::class);
+        Assert::allIsInstanceOf($shipments, UnitRefundInterface::class);
```
GSadee, line 32: Typehints of variables `$units` and `$shipments` and getter methods should be also adjusted

Nek-: What do you mean?

GSadee (approved):

GSadee:
Thank you, Maxime! :1st_place_medal:

PR #216, opened 2020-06-02
Refund list view fails if CreditMemo entity has been overridden

luca-rath:


src/Resources/config/app/config.yml
```diff
@@ -62,7 +62,7 @@ sylius_grid:
             driver:
                 name: doctrine/orm
                 options:
-                    class: Sylius\RefundPlugin\Entity\CreditMemo
+                    class: 'expr:parameter("sylius_refund.model.credit_memo.class")'
```
lchrusciel, line 65:
```suggestion
                    class: %sylius_refund.model.credit_memo.class%
```

src/Resources/config/app/config.yml
```diff
@@ -62,7 +62,7 @@ sylius_grid:
             driver:
                 name: doctrine/orm
                 options:
-                    class: Sylius\RefundPlugin\Entity\CreditMemo
+                    class: %sylius_refund.model.credit_memo.class%
```
Nek-, line 65: AFAIK this parameter is never set inside the plugin.

luca-rath: I guess it's the SyliusResourceBundle, which sets this parameter, but I'm not sure about that. But the parameter is definitely available in the container.

Nek-: I guess no? https://github.com/Sylius/SyliusResourceBundle/search?q=sylius_refund.model.credit_memo.class&unscoped_q=sylius_refund.model.credit_memo.class

luca-rath: Magic happens here https://github.com/Sylius/SyliusResourceBundle/blob/3464579ed05b199fcd8bc12e5444b36d65d6a4a5/src/Bundle/DependencyInjection/Driver/AbstractDriver.php#L48 🙂

Zales0123 (approved):

Zales0123:
Thank you, Luca, and sorry for makinging it that long 🖖

luca-rath:
No problem, mostly a little reminder is all it takes 🙂 thanks for merging!

PR #220, opened 2020-08-26
add OnlineRefundException for online refund error

oallain:
Hello,

For online refund we need to allow catching different exception, like API errors ....

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,4 +12,5 @@
 
 .env.test
 
-phpspec.yml
+/phpspec.yml
+/behat.yml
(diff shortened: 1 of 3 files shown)


GSadee (commented):
Hello @oallain! To be honest, I'm not convinced to merge this PR, as the introduced exception is not thrown anywhere. I would rather recommend you to add it in your project as the customization instead of introducing in here in plugin.

oallain:
Hi @GSadee 

Yep, I introduce it in our plugin, thanks.
Regards

PR #221, opened 2020-08-26
check online refund status before completeness

oallain:
Hello,

Before completing refund we need to dispatch a message to the eventbus for online refund handler to be able to catch errors when calling the online gateway for refund.


GSadee (commented):
Hello @oallain! First of all, sorry for replying so late. However, could you explain why this change is needed? Because if I see correctly in the code of this plugin, `RefundPaymentGenerated` is generated earlier in `RefundPaymentProcessManager` after refunding units. But the action `CompleteRefundPaymentAction` is only for changing the state of a given refund. After your change, it looks like `RefundPaymentGenerated` event is dispatched twice for one refund.

oallain:
Hello @GSadee 

I just checked and we have it like this in our project, it's been a long time so I didn't have it in mind anymore.

I close this PR.

PR #228, opened 2020-11-13
Fix Doctrine/Migrations configuration

Zales0123:
The same as https://github.com/Sylius/InvoicingPlugin/pull/195

Without this change, migrations are generated into the plugin's directory rather than `src/Migrations`.


GSadee (approved):

GSadee:
Thanks, Mateusz! :tada:

PR #232, opened 2020-11-25
Drop unused order state-machine transtion to fully refunded

diimpp:
`order->state = fully_refunded` is defined in the plugin, but never used by anything.

Besides, if such transition should exist, then it should be named `refunded` as partional refund state is called `partially_refunded`.


GSadee:
Hello @diimpp! If I see correctly, this transition is used by `OrderFullyRefundedStateResolver`, which is called in `RefundPaymentProcessManager` after `UnitsRefunded` event. So, removing this transition could break the current behaviour and some end applications. Because of that, I'm closing this PR for now, but feel free to reopen it or comment here if you still think that this change is valid.

diimpp:
@GSadee this state is not used by `OrderFullyRefundedStateResolver`, name of this resolver is misleading and it operates on `OrderPaymentState`, not `OrderState`.

https://github.com/Sylius/RefundPlugin/search?q=fully_refunded

`OrderState:fully_refunded` is an artifact from old big refactor at https://github.com/Sylius/RefundPlugin/pull/123 and not used anywhere in the app. This state in unreachable by the app.

GSadee:
Ok, my mistake, I had a wrong look at the interface this transition is going from.

GSadee (approved):
@diimpp could you add some note in UPGRADE file about removing this state and transition?

lchrusciel (approved):

diimpp:
Thanks @GSadee !

GSadee:

PR #234, opened 2020-12-09
[UI/UX] Refund float price

TheGrimmChester:
Use number type field instead of the text type. 
The web browser will send the value formatted correctly if we use 8.5 or 8,5. 
Right now on master if we put 8,5, only 8.0 is refunded but with this fix, the decimals seperated by comma are taken into account.


Jibbarth (approved):

diimpp:
It's possible to specify min/max over there as well. What do you think? Works alright on my project.

oallain (approved):
Possible improvement add 0 to min value.

GSadee (approved):

GSadee:
Thank you, Nicolas! :tada:

PR #235, opened 2020-12-10
Up Symfony Messenger requirement

loevgaard:
No reason to restrict the Symfony Messenger component

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
         "myclabs/php-enum": "^1.7",
         "sylius/resource-bundle": "^1.6",
         "sylius/sylius": "^1.7",
-        "symfony/messenger": "^4.4"
+        "symfony/messenger": "^4.4 || ^5.0"
     },
     "require-dev": {
         "behat/behat": "^3.6",


stloyd (approved):

Zales0123 (approved):
I'm wondering shouldn't we explicitly specify messenger version in our builds to make a conscious check of this compatibility 🤔 Still, as far as I see, on some builds 4.4 version is installed, on some 5.2, so it should be safe enough to go 🖖

Zales0123 (changes requested):
OK, in fact, it's not that easy :D We didn't require symfony/messenger on Sylius 1.7, but we do on Sylius 1.8... and we require 4.4 version :/ which means, we should wait for the support of this version on `sylius/sylius` itself

loevgaard:
@Zales0123: I don't see why we can't have a `^4.4 || ^5.0` on this requirement. Why would that ever conflict with `sylius/sylius`?

Zales0123:
The conflict is not explicit, we still would be able to use symfony/messenger:^4.4, but we will never be allowed to use ^5.0 with Sylius 1.8 (which, btw, is the only actively supported version now).

Zrzut ekranu 2020-12-18 o 12 53 43

That's why we would not be able to test the build with symfony/messenger:^5.0 and be sure it works 🚀

loevgaard:
I understand, Mateusz :| Hoping for SF5 support 🤞

GSadee:
Closing in favour of https://github.com/Sylius/RefundPlugin/pull/254

PR #245, opened 2021-01-14
Remove final tags from resource entities

### LINKED ISSUE #207: RefundPayment is tagged as final while its content depend on the domain
Hello guys,

The refund payment may be complex but you tagged the RefundPayment entity as final. It means that you cancel completely the possibility to use your code for payment refund if the domain needs more than just those data.

It also means that it makes no sense to have a factory building it since it cannot be different than the expected one.

Maybe I'm wrong... And if so, the architecture needs more explanation in documentation :open_mouth: . Otherwise I suggest to just remove this final comment.

GSadee:
Fixes #207


Zales0123 (approved):

lchrusciel (approved):

lchrusciel:
Thank you, Grzegorz! :tada:

PR #246, opened 2021-01-18
Upgrade to Sylius 1.9 with Symfony 5

Tomanhez:


src/Resources/config/services/command_bus.xml
```diff
@@ -10,21 +10,21 @@
             <argument type="service" id="sylius_refund_plugin.event_bus" />
             <argument type="service" id="sylius.repository.order" />
             <argument type="service" id="Sylius\RefundPlugin\Validator\RefundUnitsCommandValidator" />
-            <tag name="messenger.message_handler" />
+            <tag name="messenger.message_handler" bus="sylius_default.bus"/>
```
lchrusciel, line 13: Shouldn't this be changed to 
```suggestion
            <tag name="messenger.message_handler" bus="sylius_refund_plugin.command_bus"/>
```
to keep current behavior? But if we want to migrate to `sylius_default.bus` (which a good decision IMHO), we can remove `sylius_refund_plugin.command_bus` config entirely

src/Resources/config/app/config.yml
```diff
@@ -29,9 +29,14 @@ sylius_mailer:
 framework:
     messenger:
         buses:
-            sylius_refund_plugin.command_bus: ~
+            sylius_refund_plugin.command_bus:
+                middleware:
+                    - 'validation'
+                    - 'doctrine_transaction'
             sylius_refund_plugin.event_bus:
-                default_middleware: allow_no_handlers
```
lchrusciel, line 34: Probably should stay

lchrusciel:
Will be covered by #254

PR #249, opened 2021-01-20
Signature doesn't allow null, casting to string

loevgaard:
The signature of the orderRefundsListAvailabilityChecker doesn't allow null

### DIFF
--- a/src/Menu/OrderShowMenuListener.php
+++ b/src/Menu/OrderShowMenuListener.php
@@ -22,7 +22,7 @@ final class OrderShowMenuListener
         $menu = $event->getMenu();
         $order = $event->getOrder();
 
-        if ($this->orderRefundsListAvailabilityChecker->__invoke($order->getNumber())) {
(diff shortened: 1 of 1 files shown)


src/Menu/OrderShowMenuListener.php
```diff
@@ -22,7 +22,7 @@ public function addRefundsButton(OrderShowMenuBuilderEvent $event): void
         $menu = $event->getMenu();
         $order = $event->getOrder();
 
-        if ($this->orderRefundsListAvailabilityChecker->__invoke($order->getNumber())) {
+        if ($this->orderRefundsListAvailabilityChecker->__invoke((string) $order->getNumber())) {
```
GSadee, line 25: Can you explain in what case do you have null as an order number?

lchrusciel: It should not be possible to have an order without a number. I would rather throw an exception if the number is null

GSadee:
I'm closing that PR as it should not be possible to have an order without a number.

Issue #229, opened 2020-11-19
[Bug] CreditMemo Pdf translation uses user locale instead of credit memo locale


diimpp:
CreditMemo Pdf uses `|trans` filter, which relies on current user locale, which is admin user for admin panel.

https://github.com/Sylius/RefundPlugin/blob/02d9399de33789b9dd2d9bb431fec89576540361/src/Resources/views/Download/creditMemo.html.twig#L35

Solution is to specify `creditMemo.localeCode` in each `trans` filter
```diff
--- {{ 'sylius_refund.ui.buyer'|trans }}
+++ {{ 'sylius_refund.ui.buyer'|trans([], 'messages', creditMemo.localeCode) }}
```

Personally I'm using special macro for pdf and emails like so
```twig
 {% macro trans(message, localeCode, arguments = [], domain = 'messages') %}
     {{ message|trans(arguments, domain, localeCode) }}
 {% endmacro %}
 
 {% import _self as macro %}

 {{ macro.trans('sylius_refund.ui.buyer', creditMemo.localeCode) }}:
```

Example from my debug, where credit memo is in `de`, but my admin user locale is `en`. I've replaced some `|trans` uses with `creditMemo.localeCode`, so you can see the mismatch.

![image](https://user-images.githubusercontent.com/870747/99687187-f2272c00-2ab6-11eb-972c-67f8450cd981.png)

PR #250, opened 2021-01-28
[Pdf] set translation credit memo locale

SirDomin:
issue: https://github.com/Sylius/RefundPlugin/issues/229

Admin panel locale: fr_FR, pdf is printed in english (credit memo locale)
image

### DIFF
--- a/src/Resources/views/Download/creditMemo.html.twig
+++ b/src/Resources/views/Download/creditMemo.html.twig
@@ -32,7 +32,7 @@
                             <img src="https://demo.sylius.com/assets/shop/img/logo.png" style="width:300px;">
                         </td>
                         <td>
-                            {{ 'sylius_refund.ui.issued_at'|trans }}: {{ creditMemo.issuedAt|date('Y-m-d H:i:s') }}<br>
+                            {{ 'sylius_refund.ui.issued_at'|trans([], 'messages', creditMemo.localeCode) }}: {{ creditMemo.issuedAt|date('Y-m-d H:i:s') }}<br>
                         </td>
                     </tr>
                 </table>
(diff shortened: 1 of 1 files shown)


src/Resources/views/Download/creditMemo.html.twig
```diff
@@ -32,7 +32,7 @@
                             <img src="https://demo.sylius.com/assets/shop/img/logo.png" style="width:300px;">
                         </td>
                         <td>
-                            {{ 'sylius_refund.ui.issued_at'|trans }}: {{ creditMemo.issuedAt|date('Y-m-d H:i:s') }}<br>
+                            {{ 'sylius_refund.ui.issued_at'|trans([], 'messages', creditMemo.localeCode) }}: {{ creditMemo.issuedAt|date('Y-m-d H:i:s') }}<br>
```
AdamKasp, line 35: What do you think about extraction this `creditMemo.localeCode` to a variable, instead of copying it 17 times :D?

SirDomin: i dont think if it makes any difference :D this way we know every translation gets credit memo locale code.

pamil (approved):
Looks good to me 👍 

Anyway, we can consider changing the Translator's locale before rendering this template and then bringing back the previously used locale after it is rendered. This way we can forget about this kind of issues.

GSadee (approved):

GSadee:
Thank you, @SirDomin! :1st_place_medal:

src/Resources/views/Download/creditMemo.html.twig
```diff
@@ -32,7 +32,7 @@
                             <img src="https://demo.sylius.com/assets/shop/img/logo.png" style="width:300px;">
```
diimpp, line 32: Woah, https://demo.sylius.com/assets/shop/img/logo.png

PR #251, opened 2021-01-29
Fix failing build

SirDomin:
Added conflict to api-platform 2.6.
Removed ecs due to BC break.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -63,14 +63,13 @@
         }
     },
     "conflict": {
-        "symplify/package-builder": "^8.3.25",
+        "symplify/package-builder": "8.3.25",
         "api-platform/core": "^2.6"
     },
     "scripts": {
         "analyse": [
             "@composer validate --strict",
-            "vendor/bin/phpstan.phar analyse -c phpstan.neon -l max src/",
-            "vendor/bin/ecs check src/ spec/"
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -43,7 +43,6 @@
         "phpstan/phpstan": "0.12.29",
         "phpstan/phpstan-webmozart-assert": "0.12.6",
         "phpunit/phpunit": "^8.5",
-        "sylius-labs/coding-standard": "^3.2",
```
pamil, line 46: Let's leave it there

composer.json
```diff
@@ -63,13 +62,13 @@
...
     "conflict": {
-        "symplify/package-builder": "^8.3.25"
+        "symplify/package-builder": "8.3.25",
+        "api-platform/core": "^2.6"
     },
     "scripts": {
         "analyse": [
             "@composer validate --strict",
-            "vendor/bin/phpstan.phar analyse -c phpstan.neon -l max src/",
-            "vendor/bin/ecs check src/ spec/"
```
pamil, line 72: And just remove this line

composer.json
pamil: It's not preventing us from installing anything, it's a runtime error

GSadee (approved):

GSadee:
Thanks, @SirDomin! :1st_place_medal:

PR #254, opened 2021-02-10
Allow for Sylius 1.8 & 1.9 and Symfony 4.4 & 5.2

pamil:


.github/workflows/build.yml
```diff
@@ -11,22 +11,26 @@ on:
...
 
-        name: "PHP ${{ matrix.php }}, MySQL ${{ matrix.mysql }}, Sylius ${{ matrix.sylius }}"
+        name: "Sylius ${{ matrix.sylius }}, PHP ${{ matrix.php }}, Symfony ${{ matrix.symfony }}, MySQL ${{ matrix.mysql }}"
 
         strategy:
             fail-fast: false
             matrix:
                 php: [7.4, 7.3]
+                symfony: [^4.4, ^5.2]
+                sylius: [~1.8.0, ~1.9.0@beta]
```
Tomanhez, line 23: Why we are lock on `1.8.0` ?

src/DependencyInjection/Configuration.php
```diff
@@ -14,8 +14,8 @@ final class Configuration implements ConfigurationInterface
      */
     public function getConfigTreeBuilder(): TreeBuilder
     {
-        $treeBuilder = new TreeBuilder();
-        $rootNode = $treeBuilder->root('sylius_refund_plugin');
+        $treeBuilder = new TreeBuilder('sylius_refund_plugin');
+        $rootNode = $treeBuilder->getRootNode();
 
         return $treeBuilder;
```
Tomanhez, line 20:
```suggestion
        return new TreeBuilder('sylius_refund_plugin');
```

.github/workflows/build.yml
GSadee: We are not locked on 1.8.0, `~1.8.0` means `>=1.8.0 <1.9.0`
https://getcomposer.org/doc/articles/versions.md#tilde-version-range-

lchrusciel (commented):
Just a thought. I assume the changes to the doctrine and twig namespaces are not treated as BC breaks? Are we sure, these will be used?

.github/workflows/build.yml
```diff
@@ -124,9 +136,22 @@ jobs:
...
+                name: Create test application database
                 run: |
                     (cd tests/Application && bin/console doctrine:database:create -vvv)
+
+            -
+                name: Exclude duplicated migrations
+                if: matrix.sylius != '~1.8.0'
+                run: |
+                    (cd tests/Application && bin/console doctrine:migration:sync-metadata-storage)
+                    (cd tests/Application && bin/console doctrine:migrations:version 'Sylius\RefundPlugin\Migrations\Version20201130071338' --add)
```
pamil, line 148: Could we somehow not execute them if we detect that those changes are already done? Currently, this would require to add those three migrations for every fresh plugin installation with Sylius 1.9.

pamil (commented):
Looks good, I'd approve it if I could :)

.github/workflows/build.yml
GSadee: I don't know any way not to execute migrations if changes are already done. But I agree that it doesn't look as it should be and could be quite inconvenient during installation with Sylius 1.9. One of my thought was to move that duplicated migrations to `etc` directory and use them only for 1.8 (I've done sth like this in one of the commit: https://github.com/Sylius/RefundPlugin/pull/254/commits/6ba5b4502ad9c900a97de9d25da8ef28e720b014, but it didn't work as I wanted)

GSadee: I've added if statement to these migrations not to execute them if changes are already done

lchrusciel (approved):

GSadee:
Thanks, Kamil! :1st_place_medal:

PR #260, opened 2021-03-02
Fix OrderItemsTaxesApplicator with removed units

### LINKED ISSUE #259: OrderItemsTaxesApplicator breaks when items are removed
When items are removed from a Doctrine ArrayCollection, the indexes leave holes.

`OrderItemsTaxesApplicator` iterates over the items using an index, causing access to unset array keys:

https://github.com/Sylius/RefundPlugin/blob/7016d49d80c68b4ac888ad074946cdf3927c192d/src/TaxesApplicator/OrderItemsTaxesApplicator.php#L79

A simple fix would be to either use `$item->getUnits()->getValues()` or (less elegant) use a counter variable as seen in Sylius 1.9.

This is critical in my opinion, as it breaks all setups with RefundPlugin, so possibly also Plus.

aleho:
This fixes an undefined offset error caused by holes in ArrayCollection keys after removing units.

Fixes #259


GSadee (approved):

GSadee:
Thanks, Alexander! :tada:

PR #264, opened 2021-03-10
Taxes displayed properly on credit memo

AdamKasp:
Show Page:
Screenshot 2021-03-13 at 11 59 36

PDF:
Screenshot 2021-03-13 at 12 00 58

TODO:
- [x] implement behat context
- [x] fix row with shipment tax value
- [x] add subsums for net values and tax value
- [x] refactor PDF view


src/Converter/ShipmentLineItemsConverter.php
```diff
@@ -50,14 +50,31 @@ private function convertUnitRefundToLineItem(ShipmentRefund $shipmentRefund): Li
...
+            $grossValue,
+            $taxAmount,
+            $taxRate
         );
     }
+
+    private function getTaxRateAmount(AdjustmentInterface $adjustment): ?float
+    {
+        if (key_exists('taxRateAmount', $adjustment->getDetails())) {
+            return $adjustment->getDetails()['taxRateAmount'];
```
GSadee, line 75: Should not this value be multiplied by 100?

src/Provider/TaxRateProvider.php
```diff
@@ -5,26 +5,20 @@
...
-final class LabelBasedTaxRateProvider implements TaxRateProviderInterface
+final class TaxRateProvider implements TaxRateProviderInterface
 {
     public function provide(OrderItemUnitInterface $orderItemUnit): ?string
     {
         /** @var Collection|AdjustmentInterface[] $taxAdjustments */
         $taxAdjustments = $orderItemUnit->getAdjustments(AdjustmentInterface::TAX_ADJUSTMENT);
 
-        if ($taxAdjustments->isEmpty()) {
+        if ($taxAdjustments->isEmpty() || !key_exists('taxRateAmount', $taxAdjustments->first()->getDetails())) {
```
lchrusciel, line 18: Should we refactor it in some way, to handle more than one tax adjustment? Or trigger an exception if there is more than one?

src/Converter/ShipmentLineItemsConverter.php
```diff
@@ -50,14 +50,31 @@ private function convertUnitRefundToLineItem(ShipmentRefund $shipmentRefund): Li
         Assert::isInstanceOf($shipment, AdjustableInterface::class);
         Assert::lessThanEq($shipmentRefund->total(), $shipment->getAdjustmentsTotal());
 
+        $grossValue = $shipmentRefund->total();
+
+        $taxAdjustment = $shipment->getAdjustments(AdjustmentInterface::TAX_ADJUSTMENT)->first();
+        $taxRate = $this->getTaxRateAmount($taxAdjustment) . '%';
```
lchrusciel, line 56: Below you are multiplying the tax rate by 100(see attached image), but not here
image

Which solution is proper? And also, as this logic is so crucial, can we extract another service for it?

src/Converter/ShipmentLineItemsConverter.php
```diff
@@ -50,14 +50,31 @@ private function convertUnitRefundToLineItem(ShipmentRefund $shipmentRefund): Li
...
+            $netValue,
+            $grossValue,
+            $netValue,
+            $grossValue,
+            $taxAmount,
+            $taxRate
         );
     }
+
+    private function getTaxRateAmount(AdjustmentInterface $adjustment): ?float
```
lchrusciel, line 72: The calculation itself could be extracted as well.

Also, taking into account one of the usages of this function:
```php
$taxAmount = (int) ($grossValue * $this->getTaxRateAmount($taxAdjustment));
```
I'm not sure, we should return `null` from this function. In the example above it will be casted to 0. Won't it be better to always return float, sometimes it will be just 0? On the other hand, do we want to inform users of this function that tax rate was not found?

src/Provider/TaxRateProvider.php
AdamKasp: I vote for triggering an exception, IMO there is no possibility to have 2 tax adjustments on one order item( @CoderMaggie what do you think about it?)

src/Converter/ShipmentLineItemsConverter.php
AdamKasp: I've extracted it to a new service. I multiply it when I need string with taxRate, I left it as a float when i need `taxRate amount`

src/Provider/TaxRateProvider.php
CoderMaggie: As it is possible in the backend to have more than one tax adjustment, then we should handle all of them, not just one. 
Even if it is currently not possible to apply more tax adjustments in the core system, one might have that customized this way.

spec/Provider/TaxRateAmountProviderSpec.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+declare(strict_types=1);
+
+namespace spec\Sylius\RefundPlugin\Provider;
+
+use PhpSpec\ObjectBehavior;
+use Sylius\RefundPlugin\Entity\AdjustmentInterface;
+use Sylius\RefundPlugin\Provider\TaxRateAmountProviderInterface;
+
```
arti0090, line 10: unneded empty line

spec/Provider/TaxRateAmountProviderSpec.php
```diff
@@ -0,0 +1,36 @@
...
+        $this->shouldImplement(TaxRateAmountProviderInterface::class);
+    }
+
+    function it_provides_a_tax_rate_amount_from_tax_adjustment(AdjustmentInterface $adjustment): void
+    {
+        $adjustment->getDetails()->willReturn(['taxRateAmount' => 0.20]);
+
+        $this->provide($adjustment)->shouldReturn(0.20);
+    }
+
```
arti0090, line 25:
```suggestion

```

spec/Provider/TaxRateProviderSpec.php
```diff
@@ -26,32 +26,50 @@ function it_provides_a_tax_rate_from_tax_adjustment_label(
...
+        $this->provide($orderItemUnit)->shouldReturn(null);
     }
 
-    function it_returns_null_if_there_is_no_tax_adjustment(OrderItemUnitInterface $orderItemUnit): void
-    {
+    function it_throws_exception_if_order_item_unit_has_more_than_1_tax_adjustment(
+        OrderItemUnitInterface $orderItemUnit,
+        AdjustmentInterface $firstTaxAdjustment,
+        AdjustmentInterface $secondTaxAdjustment
+
```
arti0090, line 63:
```suggestion

```

src/Converter/ShipmentLineItemsConverter.php
```diff
@@ -50,14 +57,25 @@ private function convertUnitRefundToLineItem(ShipmentRefund $shipmentRefund): Li
...
+        $grossValue = $shipmentRefund->total();
+
+        $taxAdjustment = $shipment->getAdjustments(AdjustmentInterface::TAX_ADJUSTMENT)->first();
+
+        $taxAdjustmentAmount = $this->taxRateAmountProvider->provide($taxAdjustment);
+        $taxRate = $taxAdjustmentAmount * 100 . '%';
+        $taxAmount = (int) ($grossValue * $taxAdjustmentAmount);
+
+        $netValue = $grossValue - $taxAmount;
+
```
arti0090, line 69: I would divide it like this where there are 2 groups: 1st values are getted, and 2nd are calculated 😄 wdyt?
```suggestion
        $grossValue = $shipmentRefund->total();
        $taxAdjustment = $shipment->getAdjustments(AdjustmentInterface::TAX_ADJUSTMENT)->first();
        $taxAdjustmentAmount = $this->taxRateAmountProvider->provide($taxAdjustment);
        
        $taxRate = $taxAdjustmentAmount * 100 . '%';
        $taxAmount = (int) ($grossValue * $taxAdjustmentAmount);
        $netValue = $grossValue - $taxAmount;
```

src/Resources/views/Download/creditMemo.html.twig
```diff
@@ -101,6 +101,18 @@
...
+            <td>{{ 'sylius_refund.ui.net_total'|trans([], 'messages', creditMemo.localeCode) }}:</td>
+            <td>{{ '%0.2f'|format(creditMemo.getNetValueTotal()/100) }}</td>
+            <td>{{ creditMemo.currencyCode }}</td>
+        </tr>
+        <tr class="total">
+            <td colspan="6"></td>
+            <td>{{ 'sylius_refund.ui.tax_total'|trans([], 'messages', creditMemo.localeCode) }}:</td>
+            <td>{{ '%0.2f'|format(creditMemo.getTaxTotal()/100) }}</td>
+            <td>{{ creditMemo.currencyCode }}</td>
+        </tr>
```
arti0090, line 115:
```suggestion
        <tr class="totalNet">
            <td colspan="6"></td>
            <td>{{ 'sylius_refund.ui.net_total'|trans([], 'messages', creditMemo.localeCode) }}:</td>
            <td>{{ '%0.2f'|format(creditMemo.getNetValueTotal()/100) }}</td>
            <td>{{ creditMemo.currencyCode }}</td>
        </tr>
        <tr class="totalTax">
            <td colspan="6"></td>
            <td>{{ 'sylius_refund.ui.tax_total'|trans([], 'messages', creditMemo.localeCode) }}:</td>
            <td>{{ '%0.2f'|format(creditMemo.getTaxTotal()/100) }}</td>
            <td>{{ creditMemo.currencyCode }}</td>
        </tr>
```

In case of CSS changes, this could be usefull ( so we can change only one table), wdyt?

tests/Behat/Context/Ui/CreditMemoContext.php
```diff
@@ -218,6 +220,24 @@ public function itsTotalShouldBeInCurrency(string $total, string $currencyCode):
         Assert::same($this->creditMemoDetailsPage->getTotalCurrencyCode(), $currencyCode);
     }
 
+    /**
+     * @Then its net total should be :total
+     */
+    public function itsNetTotalShouldBe(string $total): void
+    {
+        Assert::same($this->creditMemoDetailsPage->getNetTotal(), $total);
+
```
arti0090, line 229:
```suggestion

```

tests/Behat/Context/Ui/CreditMemoContext.php
```diff
@@ -218,6 +220,24 @@ public function itsTotalShouldBeInCurrency(string $total, string $currencyCode):
...
+    }
+
+    /**
+     * @Then its tax total should be :total
+     */
+    public function itsTaxTotalShouldBe(string $total): void
+    {
+        Assert::same($this->creditMemoDetailsPage->getTaxTotal(), $total);
+
```
arti0090, line 238:
```suggestion

```

src/Resources/views/Download/creditMemo.html.twig
AdamKasp: I have a totally opposite opinion, each row have the same class because they are styled in the same way, if someone will need to customize it, then he can add a new class or specific id.

src/Provider/TaxRateProvider.php
```diff
@@ -5,26 +5,23 @@
...
 {
     public function provide(OrderItemUnitInterface $orderItemUnit): ?string
     {
         /** @var Collection|AdjustmentInterface[] $taxAdjustments */
         $taxAdjustments = $orderItemUnit->getAdjustments(AdjustmentInterface::TAX_ADJUSTMENT);
 
-        if ($taxAdjustments->isEmpty()) {
-            return null;
-        }
+        Assert::maxCount($taxAdjustments, 1, 'Every Order Item Unit Should have max 1 tax adjustment');
```
GSadee, line 19:
```suggestion
        Assert::maxCount($taxAdjustments, 1, 'Each Order Item Unit should not have more than 1 tax adjustment');
```
?
And I wonder if we shouldn't throw here a custom exception instead of using assertion, sth like `MoreThanOneTaxAdjustment`

src/Converter/ShipmentLineItemsConverter.php
```diff
@@ -50,14 +57,28 @@ private function convertUnitRefundToLineItem(ShipmentRefund $shipmentRefund): Li
...
         Assert::lessThanEq($shipmentRefund->total(), $shipment->getAdjustmentsTotal());
 
+        $grossValue = $shipmentRefund->total();
+
+        $taxAdjustments = $shipment->getAdjustments(AdjustmentInterface::TAX_ADJUSTMENT);
+
+        $taxAdjustmentAmount = 0;
+        if (count($taxAdjustments) > 0) {
+            $taxAdjustmentAmount = $this->taxRateAmountProvider->provide($taxAdjustments->first());
+        }
```
GSadee, line 67: Similar to `TaxRateProvider`, you should check if there is only one tax adjustment here

src/Converter/ShipmentLineItemsConverter.php
```diff
@@ -50,14 +57,28 @@ private function convertUnitRefundToLineItem(ShipmentRefund $shipmentRefund): Li
...
+
+        $taxRate = $taxAdjustmentAmount * 100 . '%';
+        $taxAmount = (int) ($grossValue * $taxAdjustmentAmount);
```
GSadee, line 70: I guess you can use the value from tax adjustments instead of calculating the tax amount again

src/Entity/CreditMemo.php
```diff
@@ -144,4 +144,28 @@ public function getTo(): ?ShopBillingDataInterface
...
         return $this->to;
     }
+
+    public function getNetValueTotal(): int
+    {
+        $sum = 0;
+
+        /** @var LineItemInterface $lineItem */
+        foreach ($this->getLineItems() as $lineItem) {
+            $sum = $sum + $lineItem->netValue();
```
GSadee, line 154:
```suggestion
            $sum += $lineItem->netValue();
```

src/Entity/CreditMemo.php
```diff
@@ -144,4 +144,28 @@ public function getTo(): ?ShopBillingDataInterface
...
+        return $sum;
+    }
+
+    public function getTaxTotal(): int
+    {
+        $sum = 0;
+
+        /** @var LineItemInterface $lineItem */
+        foreach ($this->getLineItems() as $lineItem) {
+            $sum = $sum + $lineItem->taxAmount();
```
GSadee, line 166:
```suggestion
            $sum += $lineItem->taxAmount();
```

src/Resources/config/services/provider.xml
```diff
@@ -50,7 +50,12 @@
 
         <service
             id="Sylius\RefundPlugin\Provider\TaxRateProviderInterface"
-            class="Sylius\RefundPlugin\Provider\LabelBasedTaxRateProvider"
+            class="Sylius\RefundPlugin\Provider\TaxRateProvider"
+        />
+
+        <service
+            id="Sylius\RefundPlugin\Provider\TaxRateAmountProviderInterface"
+             class="Sylius\RefundPlugin\Provider\TaxRateAmountProvider"
```
GSadee, line 58: Wrong indentation

tests/Behat/Context/Application/CreditMemoContext.php
```diff
@@ -150,6 +127,22 @@ public function creditMemoTotalShouldBe(int $total, string $currencyCode): void
         Assert::same($this->creditMemo->getCurrencyCode(), $currencyCode);
     }
 
+    /**
+     * @Given /^its net total should be ("[^"]+")$/
```
GSadee, line 131:
```suggestion
     * @Then /^its net total should be ("[^"]+")$/
```

tests/Behat/Context/Application/CreditMemoContext.php
```diff
@@ -150,6 +127,22 @@ public function creditMemoTotalShouldBe(int $total, string $currencyCode): void
...
+     */
+    public function itsNetTotalShouldBe(int $total): void
+    {
+        Assert::same($this->creditMemo->getNetValueTotal(), $total);
+    }
+
+    /**
+     * @Given /^its tax total should be ("[^"]+")$/
```
GSadee, line 139:
```suggestion
     * @Then /^its tax total should be ("[^"]+")$/
```

tests/Behat/Context/Ui/CreditMemoContext.php
```diff
@@ -136,16 +136,18 @@ public function itShouldContainProductWithNetValueTaxAmountAndGrossValueInCurren
     }
 
     /**
-     * @Then it should contain :quantity :shipmentName shipment with :grossValue gross value in :currencyCode currency
+     * @Then it should contain :quantity :productName shipment(s) with :netValue net value, :taxAmount tax amount and :grossValue gross value in :currencyCode currency
```
GSadee, line 139: Why did you change `shipmentName` to `productName` here?

src/Provider/TaxRateProvider.php
AdamKasp: I thought that Assert will be enough but custom exception looks better

tests/Behat/Context/Ui/CreditMemoContext.php
AdamKasp: too many copy-past :( ofc it should be `shipmentName`)

spec/Converter/ShipmentLineItemsConverterSpec.php
```diff
@@ -36,19 +39,25 @@ function it_converts_shipment_unit_refunds_to_line_items(
             ->willReturn($shippingAdjustment)
         ;
 
-        $shippingAdjustment->getLabel()->willReturn('Galaxy post');
         $shippingAdjustment->getShipment()->willReturn($shipment);
 
         $shipment->getAdjustmentsTotal()->willReturn(1000);
 
+        $shipment->getAdjustments(AdjustmentInterface::TAX_ADJUSTMENT)->willReturn( new ArrayCollection([$shippingAdjustment->getWrappedObject()]));
```
lchrusciel, line 46:
```suggestion
        $shipment->getAdjustments(AdjustmentInterface::TAX_ADJUSTMENT)->willReturn(new ArrayCollection([$shippingAdjustment->getWrappedObject()]));
```

src/Provider/TaxRateProvider.php
lchrusciel: Custom adjustment is a great idea. About the exception message in it - what would you say about: "'Each Order Item Unit must not have more than 1 tax adjustment'"?

tests/Application/src/Entity/Shipment.php
```diff
@@ -21,7 +21,7 @@
 
 /**
  * @ORM\Entity()
- * @ORM\Table(name="sylius_shipment")
+ * @ORM\Table(name="sylius_shipment")x
```
AdamKasp, line 24:
```suggestion
 * @ORM\Table(name="sylius_shipment")
```

src/Converter/ShipmentLineItemsConverter.php
```diff
@@ -45,19 +52,45 @@ private function convertUnitRefundToLineItem(ShipmentRefund $shipmentRefund): Li
...
         $shipment = $shippingAdjustment->getShipment();
         Assert::notNull($shipment);
         Assert::isInstanceOf($shipment, AdjustableInterface::class);
         Assert::lessThanEq($shipmentRefund->total(), $shipment->getAdjustmentsTotal());
 
+        $taxAdjustment = $this->getTaxAdjustment($shipment);
+        $taxAdjustmentAmount = $taxAdjustment !== null ? $taxAdjustment->getAmount() : 0;
+
+        $grossValue = $shipmentRefund->total();
+        $taxAmount = (int) ($grossValue * $taxAdjustmentAmount / $shipment->getAdjustmentsTotal());
```
lchrusciel, line 65: it may be a silly question, but why do we divide by `getAdjustmentsTotal` here?

src/Converter/ShipmentLineItemsConverter.php
```diff
@@ -45,19 +52,45 @@ private function convertUnitRefundToLineItem(ShipmentRefund $shipmentRefund): Li
...
+
+    private function getTaxAdjustment(ShipmentInterface $shipment): ?AdjustmentInterface
+    {
+        $taxAdjustments = $shipment->getAdjustments(AdjustmentInterface::TAX_ADJUSTMENT);
+        if ($taxAdjustments->isEmpty()) {
+            return null;
+        }
+
+        if ($taxAdjustments->count() > 1) {
+            throw MoreThanOneTaxAdjustment::occur();
```
lchrusciel, line 88: Should we consider the `Exception` keyword at the end of the exception name as a standard in Sylius or not?

diimpp:
image
Kind of expect to see tax rate label, not just rate over here. What do you think?

src/Converter/ShipmentLineItemsConverter.php
diimpp: +1 for `...Exception` notation

src/Converter/ShipmentLineItemsConverter.php
diimpp: It's identical to `LineItemsConverter` calc, just the way to calculate tax amount part, that corresponds to refund amount out of total unit/shipment amount.

src/Converter/ShipmentLineItemsConverter.php
GSadee: In this particular plugin, half of the exceptions have `Exception` suffix and half do not. As I remember, we wanted to avoid using this keyword but I don't insist and I don't have a good argument for either of these approaches.

src/Converter/ShipmentLineItemsConverter.php
```diff
@@ -45,19 +52,45 @@ private function convertUnitRefundToLineItem(ShipmentRefund $shipmentRefund): Li
         ;
         Assert::notNull($shippingAdjustment);
 
+        /** @var ShipmentInterface $shipment */
```
diimpp, line 55: Not exactly related to PR, but would be cool to fix as well.

```diff
--- private function convertUnitRefundToLineItem(ShipmentRefund $shipmentRefund): LineItemInterface
+++ private function convertShipmentRefundToLineItem(ShipmentRefund $shipmentRefund): LineItemInterface
```
and this by plugin notation should be `array $shipments`, but it's interfaced, so not sure what to do with it.
>    public function convert(array $units): array

src/Resources/views/Download/creditMemo.html.twig
```diff
@@ -101,6 +101,18 @@
             </tr>
         {% endfor %}
 
+        <tr class="total">
+            <td colspan="6"></td>
+            <td>{{ 'sylius_refund.ui.net_total'|trans([], 'messages', creditMemo.localeCode) }}:</td>
+            <td>{{ '%0.2f'|format(creditMemo.getNetValueTotal()/100) }}</td>
```
lchrusciel, line 107: I see this logic in several places, but I'm wondering why don't we use some kind of many formatter here, just to unify extract this division to separate template

lchrusciel (approved):

lchrusciel:
Thank you, Adam! :1st_place_medal:

src/Converter/ShipmentLineItemsConverter.php
```diff
@@ -9,17 +9,24 @@
...
 final class ShipmentLineItemsConverter implements LineItemsConverterInterface
 {
     /** @var RepositoryInterface */
     private $adjustmentRepository;
 
-    public function __construct(RepositoryInterface $adjustmentRepository)
+    /** @var TaxRateProviderInterface */
+    private $taxRateProvider;
+
+    public function __construct(RepositoryInterface $adjustmentRepository, TaxRateProviderInterface $taxRateProvider)
```
lchrusciel, line 26: This is also a bc break. Can we add a note to upgrade?

src/Converter/ShipmentLineItemsConverter.php
GSadee: Fixed in https://github.com/Sylius/RefundPlugin/pull/271

PR #273, opened 2021-03-23
Refunding multiple shipments

### LINKED ISSUE #110: Bug with tax and shipments
When calculating refunds on shipments, there are 3 bugs : 

1. Only first shipment is refunded (it fails if you have multiple shipments)
2. Tax is not applied actually
3. Order shipping promotion adjustment is not refunded

**About which tax to apply.**
For each unit you already rebuild tax from order item unit data (and not from variant/product/zone matching). That works but for shipments that's not working as long tax of each shipment is not stored per shipment adjustment but on a single order shipping adjustment, you should sum all shipment adjustement and calculate tax ratio with tax adjustment linked to the order to find which tax has been applied to a shipment.

_(That will fail in case of multiple shipments with different tax on each shipments but I'm not sure it's really a use case in the context of a single order.)_

GSadee:
Fixes #110

### DIFF
--- a/features/refunding_order_shipping_cost.feature
+++ b/features/refunding_order_shipping_cost.feature
@@ -6,24 +6,20 @@ Feature: Refunding an order shipping cost
 
     Background:
         Given the store operates on a single green channel in "United States"
-        And the store has "VAT" tax rate of 23% for "Pidgeons Services" within the "US" zone
+        And the store has "VAT" tax rate of 23% for "Shipping Services" within the "US" zone
         And the store has a product "Mr. Meeseeks T-Shirt" priced at "$10.00"
         And the store has "Galaxy Post" shipping method with "$20.00" fee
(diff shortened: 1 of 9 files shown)


tests/Behat/Context/Setup/OrderShipmentContext.php
```diff
@@ -0,0 +1,173 @@
...
+            int $quantity = 1,
+            ?ChannelInterface $channel = null
+        ): OrderInterface {
+            $order = $this->sharedStorage->get('order');
+
+            $this->addProductVariantsToOrderWithChannelPrice(
+                $order,
+                $channel ?? $this->sharedStorage->get('channel'),
+                $productVariant,
+                (int) $quantity
```
arti0090, line 139: If quantity is int from begining, should we rly cast it to int here?

arti0090 (approved):

lchrusciel (approved):

lchrusciel:
Thanks, Grzegorz! :tada:

PR #274, opened 2021-03-25
[CreditMemo] Add scenarios for viewing multiple shipments on credit memo

GSadee:


AdamKasp (approved):

tests/Behat/Context/Setup/RefundingContext.php
```diff
@@ -121,17 +121,13 @@ public function allUnitsAndShipmentFromOrderAreRefunded(
             return new OrderItemUnitRefund($unit->getId(), $unit->getTotal());
         }, $order->getItemUnits()->getValues());
 
-        /** @var ShipmentInterface $shipment */
-        $shipment = $order->getShipments()->first();
-        $shippingAdjustment = $shipment->getAdjustments(AdjustmentInterface::SHIPPING_ADJUSTMENT)->first();
+        $shipments = array_map(function(ShipmentInterface $shipment) {
+            $shippingAdjustment = $shipment->getAdjustments(AdjustmentInterface::SHIPPING_ADJUSTMENT)->first();
```
lchrusciel, line 125: Would it make sense to sum all of these adjustments or throw an exception if there are more than two?

GSadee: This is only a behat context and to be honest, there is no way to have more than one shipping adjustment on one shipment.

lchrusciel: So the only potential improvement I can still see is the usage of Doctrine API instead of the `array_map` function
`$order->getShipments()->map(fn())`

lchrusciel (approved):

lchrusciel:
Thank you, Grzegorz! :tada:

PR #283, opened 2021-04-21
[Maintenance] Upgrade PHPUnit package

GSadee:


.gitignore
```diff
@@ -10,6 +10,6 @@
 
 /tests/Application/yarn.lock
 
-.env.test
-
```
lchrusciel, line 14: Is this change needed? This file is not part of this PR

GSadee: Yup, but since I changed something in `.gitignore` file, I wanted to fix and delete a file that shouldn't be there as we have env files only in test application directory

lchrusciel:
Thank you, Grzegorz! :tada:

PR #292, opened 2021-05-21
Use creditMemoLogoPath

j92:
I noticed that the `creditMemoLogoPath` was available but not used.


GSadee (approved):

GSadee:
Thanks, Joost! :1st_place_medal:

Issue #278, opened 2021-03-30
Use resource factory to create CreditMemo in CreditMemoGenerator


pamil:
Currently, it's impossible to customise CreditMemo class.

pamil:
It's kinda hard to achieve that since we create CreditMemo through the constructor.

PR #295, opened 2021-05-24
[CreditMemo] Refactor creating CreditMemo to use a factory

GSadee:
Fixes https://github.com/Sylius/RefundPlugin/issues/278

### DIFF
--- a/spec/Factory/CreditMemoFactorySpec.php
+++ b/spec/Factory/CreditMemoFactorySpec.php
@@ -1,5 +1,14 @@
 <?php
 
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
 declare(strict_types=1);
 
 namespace spec\Sylius\RefundPlugin\Factory;
@@ -34,6 +43,7 @@ final class CreditMemoFactorySpec extends ObjectBehavior
             $currentDateTimeImmutableProvider
         );
     }
+
     function it_implements_a_credit_memo_factory_interface(): void
     {
         $this->shouldImplement(CreditMemoFactoryInterface::class);
--- a/spec/Factory/CreditMemoFactorySpec.php
+++ b/spec/Factory/CreditMemoFactorySpec.php
@@ -1,0 +1,93 @@
+<?php
+
+declare(strict_types=1);
+
+namespace spec\Sylius\RefundPlugin\Factory;
+
+use Doctrine\Common\Collections\ArrayCollection;
+use PhpSpec\ObjectBehavior;
+use Sylius\Component\Core\Model\ChannelInterface;
+use Sylius\Component\Core\Model\OrderInterface;
+use Sylius\Component\Resource\Factory\FactoryInterface;
+use Sylius\RefundPlugin\Entity\CreditMemoInterface;
+use Sylius\RefundPlugin\Entity\CustomerBillingDataInterface;
+use Sylius\RefundPlugin\Entity\LineItemInterface;
+use Sylius\RefundPlugin\Entity\ShopBillingDataInterface;
+use Sylius\RefundPlugin\Entity\TaxItemInterface;
+use Sylius\RefundPlugin\Factory\CreditMemoFactoryInterface;
+use Sylius\RefundPlugin\Generator\CreditMemoIdentifierGeneratorInterface;
+use Sylius\RefundPlugin\Generator\NumberGenerator;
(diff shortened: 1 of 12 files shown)


spec/Factory/CreditMemoFactorySpec.php
```diff
@@ -0,0 +1,93 @@
...
+        NumberGenerator $creditMemoNumberGenerator,
+        CurrentDateTimeImmutableProviderInterface $currentDateTimeImmutableProvider
+    ): void {
+        $this->beConstructedWith(
+            $creditMemoFactory,
+            $creditMemoIdentifierGenerator,
+            $creditMemoNumberGenerator,
+            $currentDateTimeImmutableProvider
+        );
+    }
```
arti0090, line 36:
```suggestion
    }
    
```

src/Factory/CreditMemoFactory.php
```diff
@@ -0,0 +1,91 @@
...
+        $this->creditMemoNumberGenerator = $creditMemoNumberGenerator;
+        $this->currentDateTimeImmutableProvider = $currentDateTimeImmutableProvider;
+    }
+
+    public function createNew(): CreditMemoInterface
+    {
+        /** @var CreditMemoInterface $creditMemo */
+        $creditMemo = $this->creditMemoFactory->createNew();
+
+        return $creditMemo;
```
arti0090, line 50:
```suggestion
        $return $this->creditMemoFactory->createNew();
```

Maybe this will work?

src/Factory/CreditMemoFactory.php
```diff
@@ -0,0 +1,91 @@
+<?php
+
+declare(strict_types=1);
```
arti0090, line 3: What about license blocks? Here and in interface? We don't use them here?

src/Resources/config/services/factory.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<container xmlns="http://symfony.com/schema/dic/services" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+    <services>
+        <defaults autowire="false" autoconfigure="false" public="true" />
+
+        <service id="Sylius\RefundPlugin\Factory\CreditMemoSequenceFactory" />
+
+        <service id="Sylius\RefundPlugin\Factory\RefundPaymentFactory">
+            <argument type="service" id="sylius.repository.payment_method"/>
```
arti0090, line 10:
```suggestion
            <argument type="service" id="sylius.repository.payment_method" />
```

src/Entity/CreditMemo.php
```diff
@@ -50,101 +49,136 @@ class CreditMemo implements CreditMemoInterface
...
+    {
+        $this->channel = $channel;
+    }
+
     public function getLineItems(): Collection
     {
         return $this->lineItems;
     }
 
+    public function setLineItems(Collection $lineItems): void
```
pamil, line 127: Could it be replaced by the adder method?

Anyway, we should create an empty collection in the constructor as well - currently, it'll crash while flushing the changes if `setLineItems` and `setTaxItems` are not called.

src/Factory/CreditMemoFactory.php
GSadee: I need to add a type hint to this variable because of static analysis

src/Resources/config/services/factory.xml
```diff
@@ -0,0 +1,26 @@
...
+        </service>
+
+        <service
+            id="sylius_refund.custom_factory.credit_memo"
```
pamil, line 14: What do you think about setting id to `Sylius\RefundPlugin\Factory\CreditMemoFactory`? The current name is not used in any place outside of this configuration and will not be public. This way we won't have to call it a "custom factory".

AdamKasp (approved):

src/Factory/CreditMemoFactory.php
GSadee: We don't have them in almost the entire project, but we probably should 🤔

src/Entity/CreditMemo.php
GSadee: I've added creating collections in constructor, but I'm not convince to changing these setters (`setTaxItems()` and `setLineItems()`) to adders as these are collections that shouldn't have too many operations

src/Factory/CreditMemoFactory.php
pamil: Should be handled by the coding standard tool.

pamil (approved):

pamil:
Thanks, Grzegorz! :tada:

Issue #91, opened 2018-10-12
Is Process Manager a suitable solution when refunding order units?


bartoszpietrzak1994:
Here's the thing:

We, as Sylius Core Team members, decided to avoid a direct relation between Credit Memo and Refund Payment entities in order to make Refund Plugin's logic possibly flexible and extendable.

Hence, we implemented two separate ProcessManagers that are listeners for UnitsRefunded event - one of them is responsible for creating Credit Memo, the other for creating Refund Payment. Technically, there is no relation between these two entities but...

Let's assume that CreditMemoProcessManger is the first listener to receive an event. When the process crashes, one can always stop event's propagation in `catch` clause and avoid processing it by RefundPaymentProcessManager. 

However, since both services are Process Managers and there is no rollback mechanism, there's nothing we can do when an error occurs in the last manager. It can result in two possible inconsistent states when there's a Credit Memo with no Refund Payment or a Refund Payment without Credit Memo. 

During small office discussion we've thought about reimplementing UnitsRefunded event listeners in order to create a mechanism similar to a saga with proper rollback mechanisms implemented. 

What do you guys think?

diimpp:
While it's true, that whole process should be atomic and generate either all or nothing, I do not agree, that Refund Payment and Credit memo shouldn't me related.

Credit Memos are not generated without reason and if for some reason you don't want to add strict DB relationship, it's possible to add something like `author_id` at `Sylius\Core\Model\Image`, that can be reused in different ways.

For example, I need to implement Google Analytics refunds and supply transaction id, I may use Credit Memo number for that.

Additionally, I think Credit Memo should be generated only after Refund Payment is completed, not at the same time as Refunds are created. (Credit Memo is document for refund payment, so I don't see how they can be unrelated).

On separate note, I think Refunds should be grouped somehow per request. It should be possible to different between multiple Refunds by in which request they were done (Source of the Refund Payment and Credit Memo) as right now it's possible to know which Refunds were the source of Refund Payment and Credit Memo

GSadee:
Closing as it has been fixed in #296

PR #296, opened 2021-05-24
Always create RefundPayment after CreditMemo

Zales0123:
Partially solves https://github.com/Sylius/RefundPlugin/issues/91

With the current implementation, both **CreditMemo** and **RefundPayment** are generated separately, which can results in a situation when one of them exists and the other does not. With the proposed change, we're protected from such a situation.

For sure, the solution is not perfect. The whole compensation/rollback functionality is based on the [DoctrineTransactionMiddleware](https://github.com/symfony/doctrine-bridge/blob/5.4/Messenger/DoctrineTransactionMiddleware.php#L37), which works in this case but does not have to be enough for some more complicated processes. A possible improvement of the proposed architecture would be adding the `prev/compensate/rollback` function to the `UnitsRefundedProcessStepInterface` to implement a more mature and reliable saga pattern 🖖

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,6 +161,23 @@ as shown below:
 Online refund logic should be implemented if you need it.
 As the first try for the possible customization, you can check out `Sylius\RefundPlugin\Event\UnitsRefunded` event.
 
+## Post-refunding process
+
+After units are refunded, there are multiple other processes that should be triggered. By default, after units refund, there should be **CreditMemo** and
+**RefundPayment** generated. As they're strictly coupled with each other, **RefundPayment** is always created after the **CreditMemo**. Moreover, if **RefundPayment**
+fails, related **CreditMemo** should not be created as well.
+
+`\Sylius\RefundPlugin\ProcessManager\UnitsRefundedProcessManager` service facilitates the whole process. If you want to add one or more steps to it, you should create
+a service implementing `\Sylius\RefundPlugin\ProcessManager\UnitsRefundedProcessStepInterface`, and register if with proper tag:
+
+```xml
+    <service id="App\ProcessManager\CustomAfterRefundProcessManager">
+        <tag name="sylius_refund.units_refunded.process_step" priority="0" />
+    </service>
+```
+
+Tagged services would be executed according to their priority (descending). 
+
 ## Security issues
 
 If you think that you have found a security issue, please do not use the issue tracker and do not post it publicly.
--- a/README.md
+++ b/README.md
@@ -167,13 +167,13 @@ After units are refunded, there are multiple other processes that should be trig
 **RefundPayment** generated. As they're strictly coupled with each other, **RefundPayment** is always created after the **CreditMemo**. Moreover, if **RefundPayment**
 fails, related **CreditMemo** should not be created as well.
 
-`\Sylius\RefundPlugin\ProcessManager\UnitsRefundedProcessManager` service facilitates the whole process. If you want to add one or more steps to it, you should create
-a service implementing `\Sylius\RefundPlugin\ProcessManager\UnitsRefundedProcessStepInterface`, and register if with proper tag:
+`Sylius\RefundPlugin\ProcessManager\UnitsRefundedProcessManager` service facilitates the whole process. If you want to add one or more steps to it, you should create
+a service implementing `Sylius\RefundPlugin\ProcessManager\UnitsRefundedProcessStepInterface`, and register if with proper tag:
 
-```xml
-    <service id="App\ProcessManager\CustomAfterRefundProcessManager">
-        <tag name="sylius_refund.units_refunded.process_step" priority="0" />
-    </service>
+```yaml
+    App\ProcessManager\CustomAfterRefundProcessManager:
+       tags:
+          - { name: sylius_refund.units_refunded.process_step, priority: 0 }
 ```
 
 Tagged services would be executed according to their priority (descending). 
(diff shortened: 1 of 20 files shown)


tests/Behat/Context/Setup/RefundingContext.php
```diff
@@ -153,6 +167,22 @@ public function shipmentFromOrderHasAlreadyRefundedWithPayment(
         ));
     }
 
+    /**
+     * @Given the credit memo generation is broken
+     */
+    public function theCreditMemoGenerationIsBroken(): void
+    {
+        $this->failedCreditMemoGenerator->failCreditMemoGeneration();;
```
GSadee, line 175:
```suggestion
        $this->failedCreditMemoGenerator->failCreditMemoGeneration();
```

GSadee (approved):

README.md
```diff
@@ -161,6 +161,23 @@ as shown below:
...
+fails, related **CreditMemo** should not be created as well.
+
+`\Sylius\RefundPlugin\ProcessManager\UnitsRefundedProcessManager` service facilitates the whole process. If you want to add one or more steps to it, you should create
+a service implementing `\Sylius\RefundPlugin\ProcessManager\UnitsRefundedProcessStepInterface`, and register if with proper tag:
+
+```xml
+    <service id="App\ProcessManager\CustomAfterRefundProcessManager">
+        <tag name="sylius_refund.units_refunded.process_step" priority="0" />
+    </service>
+```
```
arti0090, line 177: Is a xml a way to go here or we should add yaml 🤔  ?

features/not_seeing_refund_payments_and_credit_memos_on_admin_order_view.feature
```diff
@@ -0,0 +1,33 @@
+@refunds
+Feature: Not seeing refund payments and credit memos on admin order view
+    In order to be prevented from having incorrectly generated credit memos or refund payments
+    As an Administrator
+    I want not to see them if process fails on any step
```
arti0090, line 5: 🚀 
```suggestion
   I don't want to see them if process fails on any step
```
or
```suggestion
    I do not want to see them if process fails on any step
```

spec/ProcessManager/UnitsRefundedProcessManagerSpec.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+declare(strict_types=1);
```
arti0090, line 3: Copyright/license block?

src/ProcessManager/UnitsRefundedProcessManager.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+declare(strict_types=1);
```
arti0090, line 3: Copyright/license block?

src/ProcessManager/UnitsRefundedProcessManagerInterface.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
```
arti0090, line 3: Copyright/license block? 🙃  also next Interface

tests/Behat/Services/Factory/FailedRefundPaymentFactory.php
```diff
@@ -0,0 +1,42 @@
+<?php
+
+declare(strict_types=1);
```
arti0090, line 3: Guess what? .... license block xD

features/not_seeing_refund_payments_and_credit_memos_on_admin_order_view.feature
GSadee: To be honest, I would for the original sentence 🚀

pamil: I don't want to generate just one of them?

pamil: Not seeing them would just mean that we can hide them in UI and job's done 🎉

tests/Behat/Services/Factory/FailedRefundPaymentFactory.php
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    public function createWithData(
+        string $orderNumber,
+        int $amount,
+        string $currencyCode,
+        string $state,
+        int $paymentMethodId
+    ): RefundPaymentInterface {
+        if (file_exists(self::FAILED_FILE)) {
```
pamil, line 29: I guess we should delete that file on `__construct` to make sure that we're always starting from the same point. Eg. if the app breaks between `failRefundPaymentCreation` and `createWithData`, we're left in an incorrect state.

AdamKasp (approved):

README.md
Zales0123: Could be yaml 👌

spec/ProcessManager/UnitsRefundedProcessManagerSpec.php
Zales0123: We generally do not have them within the whole RefundPlugin 💃 could be added, but we can do it all at once in a separate PR

features/not_seeing_refund_payments_and_credit_memos_on_admin_order_view.feature
Zales0123: I thought a little bit about a goal as well, and I came up with:

```
In order to have consistent documents and payment in a refunded order
As an Administrator
I don't want to have a credit memo or refund payment generated alone
```

Feel free to comment if you see any danger in such a change 😄 🖖

tests/Behat/Services/Factory/FailedRefundPaymentFactory.php
Zales0123: Apparently, it does not work as expected :/ The constructor is called twice and removes the file unnecessarily. I've added a `@AfterScenario` hook to the context, to do the trick 🖖

GSadee (approved):

GSadee:

PR #303, opened 2021-06-01
Change mink-debug-extension and remove unnedded params

arti0090:
Lakion version of mink is not supported anymore. This can lead to issues with f.e. PHP 8.0 version.


behat.yml.dist
```diff
@@ -6,11 +6,6 @@ default:
     extensions:
         DMore\ChromeExtension\Behat\ServiceContainer\ChromeExtension: ~
 
-        Lakion\Behat\MinkDebugExtension:
-            directory: etc/build
-            clean_start: false
-            screenshot: true
```
GSadee, line 12: Shouldn't we change to use `FriendsOfBehat\MinkDebugExtension` instead of removing this part of config?

arti0090: Both ways will work, if this lines are not here they are taken from vendor/sylius.

GSadee (approved):

GSadee:
Thank you, @arti0090! :1st_place_medal:

PR #307, opened 2021-06-09
[RefundPayment] Change order number to Order relation on RefundPayment

### LINKED ISSUE #193: [RFC] It's not possible to distinguish between refund requests; Incorrect DB design?

GSadee:
Partially fixes #193

Similar to #306

### DIFF
--- a/src/Doctrine/ORM/RefundPaymentRepository.php
+++ b/src/Doctrine/ORM/RefundPaymentRepository.php
@@ -1,0 +1,31 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\RefundPlugin\Doctrine\ORM;
+
+use Sylius\Bundle\ResourceBundle\Doctrine\ORM\EntityRepository;
+use Sylius\RefundPlugin\Repository\RefundPaymentRepositoryInterface;
+
+class RefundPaymentRepository extends EntityRepository implements RefundPaymentRepositoryInterface
+{
+    public function findByOrderNumber(string $orderNumber): array
+    {
+        return $this->createQueryBuilder('o')
+            ->innerJoin('o.order', 'ord')
+            ->andWhere('ord.number = :orderNumber')
+            ->setParameter('orderNumber', $orderNumber)
+            ->getQuery()
+            ->getResult()
+        ;
+    }
+}
--- a/src/ProcessManager/RefundPaymentProcessManager.php
+++ b/src/ProcessManager/RefundPaymentProcessManager.php
@@ -14,6 +14,8 @@ declare(strict_types=1);
 namespace Sylius\RefundPlugin\ProcessManager;
 
 use Doctrine\ORM\EntityManagerInterface;
+use Sylius\Component\Core\Model\OrderInterface;
+use Sylius\Component\Core\Repository\OrderRepositoryInterface;
 use Sylius\RefundPlugin\Entity\RefundPaymentInterface;
 use Sylius\RefundPlugin\Event\RefundPaymentGenerated;
 use Sylius\RefundPlugin\Event\UnitsRefunded;
@@ -21,6 +23,7 @@ use Sylius\RefundPlugin\Factory\RefundPaymentFactoryInterface;
 use Sylius\RefundPlugin\Provider\RelatedPaymentIdProviderInterface;
 use Sylius\RefundPlugin\StateResolver\OrderFullyRefundedStateResolverInterface;
 use Symfony\Component\Messenger\MessageBusInterface;
+use Webmozart\Assert\Assert;
 
 final class RefundPaymentProcessManager implements UnitsRefundedProcessStepInterface
 {
@@ -33,6 +36,9 @@ final class RefundPaymentProcessManager implements UnitsRefundedProcessStepInter
     /** @var RefundPaymentFactoryInterface */
     private $refundPaymentFactory;
 
+    /** @var OrderRepositoryInterface */
+    private $orderRepository;
+
     /** @var EntityManagerInterface */
     private $entityManager;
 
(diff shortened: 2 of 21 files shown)


src/ProcessManager/RefundPaymentProcessManager.php
```diff
@@ -43,20 +49,26 @@ public function __construct(
...
         $this->refundPaymentFactory = $refundPaymentFactory;
+        $this->orderRepository = $orderRepository;
         $this->entityManager = $entityManager;
         $this->eventBus = $eventBus;
     }
 
     public function next(UnitsRefunded $unitsRefunded): void
     {
+        /** @var OrderInterface|null $order */
+        $order = $this->orderRepository->findOneByNumber($unitsRefunded->orderNumber());
```
GSadee, line 67: I wonder if I should find the order here or in `RefundPaymentFactory` 🤔 Especially that we are already finding the payment method by its id in the factory. Wdyt?

Zales0123: In my opinion, we should pass to the factory as final data as possible, to allow it to create the desired object _and do nothing more_ 🚀

src/Resources/config/admin_routing.yml
```diff
@@ -59,8 +59,8 @@ sylius_refund_order_refund_payment_list:
         _sylius:
             template: "@SyliusRefundPlugin/Order/Admin/RefundPayment/list.html.twig"
             repository:
-                method: findByOrderNumber
-                arguments: $orderNumber
+                method: findByOrder
+                arguments: expr:service('sylius.repository.order').findOneByNumber($orderNumber)
```
Zales0123, line 63: I'm not sure about this approach. We're calling the repository to retrieve the object, to pass it to the same repository but another function 💃 maybe we could just introduce a custom repository method that would take the order number and join the orders' table in a query builder?

src/Resources/config/services/providers.xml
```diff
@@ -53,7 +53,6 @@
             id="Sylius\RefundPlugin\Provider\RelatedPaymentIdProviderInterface"
             class="Sylius\RefundPlugin\Provider\DefaultRelatedPaymentIdProvider"
         >
-            <argument type="service" id="sylius.repository.order" />
         </service>
```
Zales0123, line 56: So it could be shorten:
```xml
<service
   id="Sylius\RefundPlugin\Provider\RelatedPaymentIdProviderInterface"
   class="Sylius\RefundPlugin\Provider\DefaultRelatedPaymentIdProvider"
/>
```

src/Resources/config/admin_routing.yml
GSadee: It is not the same repository, I'm calling the order repository to retrieve the order, to pass it to the refund payment repository.
However, introducing a custom repository method is ok for me, I wonder about that, but the above solution was just simpler, but I agree that it was probably the worse option due to performance (2 calls to db instead of 1 with additional join).

Zales0123: Yeah, indeed, these are 2 different repositories xD Sorry, I had a hallucination 💃

Tomanhez (approved):

SirDomin (approved):

src/Doctrine/ORM/RefundPaymentRepository.php
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\RefundPlugin\Doctrine\ORM;
+
+use Sylius\Bundle\ResourceBundle\Doctrine\ORM\EntityRepository;
+use Sylius\RefundPlugin\Repository\RefundPaymentRepositoryInterface;
+
+class RefundPaymentRepository extends EntityRepository implements RefundPaymentRepositoryInterface
```
Zales0123, line 19: Should we (similar to factories) provide trait that should be implemented in the repository? But I suppose it's not a case for this change but for all the repositories and plugin 💃

src/Doctrine/ORM/RefundPaymentRepository.php
```diff
@@ -0,0 +1,31 @@
...
+{
+    public function findByOrderNumber(string $orderNumber): array
+    {
+        return $this->createQueryBuilder('o')
+            ->innerJoin('o.order', 'ord')
```
Zales0123, line 24: `ord` 😱 Is it because `order` is a banned name? 😄 maybe then `refundPaymentOrder`?

Zales0123 (approved):

Zales0123:
Thanks, Grzegorz! :tada:

src/Doctrine/ORM/RefundPaymentRepository.php
GSadee: Yup, that is the reason 😃 And to be honest, it is a copy-pasted shortcut from Sylius, but I agree that `refundPaymentOrder` looks better 😃

src/Doctrine/ORM/RefundPaymentRepository.php
GSadee: I think so, but this is the plugin, that defines RefundPayment, so creating a repository without traits makes sense here.

PR #309, opened 2021-06-10
Order refund admin show rebuild

Tomanhez:
New:
Screenshot 2021-06-10 at 15 05 05


Zales0123 (commented):
Also button styling is quite poor :( we should make it more consistent with buttons on regular Sylius' payments 🐎

tests/Application/templates/bundles/SyliusAdminBundle/Order/Show/_payments.html.twig
```diff
@@ -0,0 +1,18 @@
+{% if order.hasPayments %}
```
GSadee, line 1: We should link these templates with `src/Resources/templates/bundles/*`

GSadee (approved):

GSadee:

Issue #279, opened 2021-04-01
Is there a way to override the RefundType without override every service using it ?


Kazadri:
In the the project I'm working on, we had to add "service" to order item but they need to be refundable. For this case I overrided the `RefundType` to add a service type but since the class is use in multiple service I need to override all of them just to change the use.

Is there an other way to add a refund type so I don't need to override service like `UnitRefundedTotalProvider` or `UnitRefundingAvailabilityChecker` ?

PR #311, opened 2021-06-14
Add possibility to customize RefundType

Tomanhez:
Fix for: #279

How to custom RefundType:

After added new RefundType and RefundTypeInterface with new features:
```
<?php

declare(strict_types=1);

namespace App\Entity;

use Sylius\RefundPlugin\Model\RefundType as BaseRefundType;

final class RefundType extends BaseRefundType implements RefundTypeInterface
{
    public static function customUnit(): RefundTypeInterface
    {
        return new self(self::CUSTOM_UNIT);
    }
}
```

```
<?php

declare(strict_types=1);

namespace App\Entity;

use Sylius\RefundPlugin\Model\RefundTypeInterface as BaseRefundTypeInterface;

interface RefundTypeInterface extends BaseRefundTypeInterface
{
    public const CUSTOM_UNIT = 'custom_unit';

    public static function customUnit(): self;
}
```

You have to:

- Overwrite `RefundEnumType` and createType() method for creating new `RefundType`:

```
<?php

declare(strict_types=1);

namespace App\Entity\Type;

use App\Entity\RefundType;
use Sylius\RefundPlugin\Entity\Type\RefundEnumType as BaseRefundEnumType;
use Sylius\RefundPlugin\Model\RefundTypeInterface;

class RefundEnumType extends BaseRefundEnumType
{
    protected function createType($value): RefundTypeInterface
    {
        return new RefundType($value);
    }
}
```

Overwrite `sylius_refund.refund_enum_type` key with new `RefundEnumType`: `App\Entity\Type\RefundEnumType`
and `sylius_refund.refund_type` key with new `RefundType`: ``App\Entity\Type\RefundType``

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -80,6 +80,7 @@
 1. `sylius_refund_plugin.block_event_listener.account.order_show` and `sylius_refund_plugin.block_event_listener.order_show.credit_memos`
     listeners have been replaced by `sylius_ui` configuration
 
+<<<<<<< HEAD
 1. `Sylius\RefundPlugin\Factory\RefundPaymentFactoryInterface` and `Sylius\RefundPlugin\Factory\RefundPaymentFactory`
     service definitions have been removed and replaced by default resource factory `sylius_refund.factory.refund_payment`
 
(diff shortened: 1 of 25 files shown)


src/Entity/Type/RefundEnumType.php
```diff
@@ -30,31 +31,37 @@ public function getSQLDeclaration(array $fieldDeclaration, AbstractPlatform $pla
         return 'VARCHAR(256)';
     }
 
-    public function convertToPHPValue($value, AbstractPlatform $platform): RefundType
+    public function convertToPHPValue($value, AbstractPlatform $platform): RefundTypeInterface
     {
-        if (!RefundType::isValid($value)) {
+        if ($value instanceof RefundType && !$value::isValid($value)) {
```
GSadee, line 36: We should probably check here `RefundTypeInterface` instead of `RefundType`

Tomanhez: Yes but interface doesn't have `isValid`, and is problem with code analyse. I can check `$value instanceof RefundTypeInterface && $value instanceof Enum` but checking with `RefundType` work because we need overwrite this class when we create new `RefundType` and `instanceof` see this object as old `RefundType` because is his parent

src/Provider/RemainingTotalProvider.php
```diff
@@ -60,9 +62,9 @@ public function getTotalLeftToRefund(int $id, RefundType $type): int
         return $unitTotal - $refundedTotal;
     }
 
-    private function getRefundUnitTotal(int $id, RefundType $refundType): int
+    private function getRefundUnitTotal(int $id, RefundTypeInterface $refundType): int
     {
-        if ($refundType->equals(RefundType::orderItemUnit())) {
+        if ($refundType instanceof Enum && $refundType->equals(RefundType::orderItemUnit())) {
```
Zales0123, line 67: It would probably be worth introducing `equals` method from `Enum` class on `RefundTypeInterface`. This way, we would not have to check everywhere is `RefundType` an enum or use the class instead of the abstraction

UPGRADE.md
```diff
@@ -73,6 +73,142 @@
...
+            RepositoryInterface $refundPaymentRepository,
+    +       RefundTypeFactoryInterface $refundTypeFactory
+        ) {
+            ...
+        }
+    ```
+
+1. The `__invoke` method of `Sylius\RefundPlugin\Checker\UnitRefundingAvailabilityChecker` and `Sylius\RefundPlugin\Checker\UnitRefundingAvailabilityCheckerInterface` has been changed:
+
+    ```diff
```
Zales0123, line 94: I'm not 100% sure are these diffs useful enough to justify a few additional lines of a file in each point 😄 Maybe simple

```md
`Sylius\RefundPlugin\Checker\UnitRefundingAvailabilityCheckerInterface::__invoke` second parameter has been changed from `RefundType $refundType` to `RefundTypeInterface $refundType`
```

would do the trick? I'm not sure, do we have to mention changes on the class itself, as it's obvious because of interface change :)

pamil: In these cases, we can just write that we replaced `RefundType` with `RefundTypeInterface` in the given list of classes.

src/Entity/Type/RefundEnumType.php
```diff
@@ -30,31 +31,37 @@ public function getSQLDeclaration(array $fieldDeclaration, AbstractPlatform $pla
...
         }
 
-        throw ConversionException::conversionFailed($value, 'sylius_refund_refund_type');
+        throw ConversionException::conversionFailed((string) $value, 'sylius_refund_refund_type');
+    }
+
+    /** @param RefundTypeInterface|string $value */
+    protected function createType($value): RefundTypeInterface
+    {
+        return new RefundType($value);
```
pamil, line 65: This way the implementation is still coupled to `RefundType` instead of `RefundTypeInterface` and its factory.

src/Model/RefundTypeInterface.php
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\RefundPlugin\Model;
+
+interface RefundTypeInterface
```
pamil, line 16: Could we add `valid()` method here to take care of https://github.com/Sylius/RefundPlugin/pull/311/files#r652550575?

pamil (changes requested):

src/Entity/Type/RefundEnumType.php
Tomanhez: Good idea but this class in not a service, Do you have any idea how to inject factory with param as `RefundType` FQCN?

src/Model/RefundTypeInterface.php
```diff
@@ -0,0 +1,42 @@
...
+
+    /**
+     * @psalm-param mixed $variable
+     */
+    public function equals($variable = null): bool;
+
+    /**
+     * @psalm-param mixed $value
+     */
+    public static function isValid($value): bool;
```
Zales0123, line 34: 😱 static methods in interfaces 😱 I didn't even know we can do that 😄 Nevertheless, it's definitely not proper coding, static methods are meant to be called without created object, but they need a function body for that 💃 Can we change them to be not static? I know they're done this way in the `Enum` class.

src/Entity/Type/RefundEnumType.php
Zales0123: Sadly, we could not be able to do that :( If we're forced to implement it this way, I would add information on how to extend this type in the application that customizes `RefundType` class 🖖

src/Model/RefundTypeInterface.php
Tomanhez: Is not possible to call static method from parent in any not static method

Zales0123: Indeed, but it does not mean we should go with such an abomination as static methods in the interface 😄 
What if we add checking `Enum` instance in `RefundEnumType::converToPHPValue` method and remove this functions from both `RefundTypeInterface` and `RefundType`?

```
if (
    $value instanceof RefundTypeInterface &&
    $value instanceof Enum &&
    !$value::isValid($value)
) {
    // ...
}
```

This way we're sure that the object implements `RefundTypeInterface` (which is required), is an enum (which is also required) and is a valid enum 🚀 

cc @pamil as you were also a part of this discussion

Tomanhez: Its my first version, before I added this methods. For me both are good

Zales0123: As far as I remember, the first version was `$value instanceof RefundType && !$value::isValid($value)`, which was not using the interface - that was a problem 🚀

src/Entity/Type/RefundEnumType.php
Zales0123: Is it mentioned anywhere? If not, I would keep it open :)

src/Model/RefundType.php
```diff
@@ -30,4 +26,12 @@ public static function shipment(): self
     {
         return new self(self::SHIPMENT);
     }
+
+    /**
+     * @psalm-return mixed
+     */
+    public function getValue()
```
Zales0123, line 33: Do we need it?

src/Provider/RemainingTotalProvider.php
Zales0123: I believe we still can have this equals method on the interface? Or is it a problem, because it's a final function on the `Enum` class?

Zales0123:
Rebase needed, a few more comments and it's good to go 🖖

src/Model/RefundType.php
Tomanhez: Its using twice times in `RefundEnumType.php`

src/Entity/Type/RefundEnumType.php
Tomanhez: Its mentioned in PR description and new task for creating docs is ready

src/Provider/RemainingTotalProvider.php
Tomanhez: Its final function and it is not possible to extend it on child class

src/Model/RefundType.php
GSadee: Maybe we should cast the value to string?

src/Resources/config/services/factories.xml
```diff
@@ -12,6 +12,16 @@
...
 
+        <service id="Sylius\RefundPlugin\Factory\RefundTypeFactoryInterface" class="Sylius\RefundPlugin\Factory\RefundTypeFactory">
+            <argument type="string">%sylius_refund.refund_type%</argument>
+        </service>
+        <service id="Sylius\RefundPlugin\Factory\RefundPaymentFactoryInterface" class="Sylius\RefundPlugin\Factory\RefundPaymentFactory">
+            <argument type="service" id="sylius.repository.payment_method" />
+        </service>
+        <service id="Sylius\RefundPlugin\Factory\RefundPaymentFactory" alias="Sylius\RefundPlugin\Factory\RefundPaymentFactoryInterface">
+            <deprecated>The "%alias_id%" service alias is deprecated and will be removed in RefundPlugin 1.0, use Sylius\RefundPlugin\Factory\RefundPaymentFactoryInterface instead.</deprecated>
+        </service>
```
GSadee, line 23: Why did you add these services here?
```suggestion
```

GSadee (commented):
There is unfortunatelly new class that uses `RefundType` and should be also refactored: `Sylius\RefundPlugin\Converter\RefundUnitsConverter`

GSadee:
Closing in favour of #323

PR #312, opened 2021-06-15
Improve dx with extending refund type

AdamKasp:


GSadee (approved):

src/Command/RefundUnits.php
```diff
@@ -17,7 +17,7 @@
 use Sylius\RefundPlugin\Model\ShipmentRefund;
 use Webmozart\Assert\Assert;
 
-final class RefundUnits
+class RefundUnits implements RefundUnitsInterface
```
Zales0123, line 20: I wonder, do we need an interface if this class is no longer final? I suppose, if we extend the class it could still be handled properly by the handler (which, nevertheless, should be overridden/decorated to use additional command data), and can still be returned by the command creator.

If we decide to introduce an interface for this command, should also introduce it for all of them 💃

Zales0123 (approved):

Tomanhez (approved):

lchrusciel (approved):

lchrusciel:
Thanks, Adam! :1st_place_medal:

PR #313, opened 2021-06-15
Change CustomerBillingData to resource and create via factory

arti0090:


SirDomin (approved):

src/Entity/CustomerBillingData.php
```diff
@@ -46,80 +46,108 @@ class CustomerBillingData implements CustomerBillingDataInterface
...
+    {
+        $this->firstName = $firstName;
+    }
+
+    public function getLastName(): string
     {
         return $this->lastName;
     }
 
-    public function fullName(): string
```
Tomanhez, line 86: We should add note in upgrade file about changed `public function fullName(): string` to `public function getStreet(): string` and all others

src/Entity/CustomerBillingData.php
```diff
@@ -46,80 +46,108 @@ class CustomerBillingData implements CustomerBillingDataInterface
...
-        $this->firstName = $firstName;
-        $this->lastName = $lastName;
-        $this->street = $street;
-        $this->postcode = $postcode;
-        $this->countryCode = $countryCode;
-        $this->city = $city;
-        $this->company = $company;
-        $this->provinceName = $provinceName;
-        $this->provinceCode = $provinceCode;
+    public function getId(): ?int
```
GSadee, line 49: I would be for not forcing the type of identifier
```suggestion
    public function getId()
```

src/Entity/CustomerBillingData.php
```diff
@@ -46,80 +46,108 @@ class CustomerBillingData implements CustomerBillingDataInterface
...
+    {
+        return $this->id;
     }
 
-    public function id(): ?int
+    public function setId(?int $id): void
```
GSadee, line 54:
```suggestion
    public function setId($id): void
```

src/Entity/CustomerBillingDataInterface.php
```diff
@@ -13,27 +13,47 @@
 
 namespace Sylius\RefundPlugin\Entity;
 
-interface CustomerBillingDataInterface
+use Sylius\Component\Resource\Model\ResourceInterface;
+
+interface CustomerBillingDataInterface extends ResourceInterface
 {
-    public function id(): ?int;
+    public function setId(?int $id): void;
```
GSadee, line 20:
```suggestion
    public function setId($id): void;
```

UPGRADE.md
```diff
@@ -23,6 +23,48 @@
...
+    -   public function postcode(): string
+    +   public function getPostcode(): string
+    -   public function city(): string
+    +   public function getCity(): string
+    -   public function company(): ?string
+    +   public function getCompany(): ?string
+    -   public function provinceName(): ?string
+    +   public function getProvinceName(): ?string
+    -   public function provinceCode(): ?string
+    +   public function getProvinceCode(): ?string
```
GSadee, line 65: I'm not sure if we should change these getters, probably the better way would be not changing them or adding new ones and leaving old methods with deprecation

arti0090: I have added deprecated annotations and leaved a note in file about upcoming changes 🙂

UPGRADE.md
```diff
@@ -23,6 +23,56 @@
...
+        }
+    ```
+
+1. The `id` method from `Sylius\RefundPlugin\Entity\CustomerBillingData` has its return type removed:
+    
+    ```diff
+    -   public function id(): ?int;
+    +   public function id();
+    ```
+
```
GSadee, line 49:
```suggestion
```

GSadee (approved):

GSadee:
Thank you, @arti0090! :1st_place_medal:

PR #315, opened 2021-06-16
Shop billing data as resource

SirDomin:


src/Factory/ShopBillingDataFactory.php
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\RefundPlugin\Factory;
+
+use Sylius\Component\Resource\Factory\FactoryInterface;
+use Sylius\RefundPlugin\Entity\ShopBillingDataInterface;
+
+class ShopBillingDataFactory implements ShopBillingDataFactoryInterface
```
lchrusciel, line 19: could be finalized 
```suggestion
final class ShopBillingDataFactory implements ShopBillingDataFactoryInterface
```

lchrusciel (approved):

Tomanhez (approved):

src/Entity/ShopBillingData.php
```diff
@@ -37,19 +37,73 @@ class ShopBillingData implements ShopBillingDataInterface
...
-        ?string $street,
-        ?string $city,
-        ?string $postcode
-    ) {
+    public function getId(): ?int
+    {
+        return $this->id;
+    }
+
+    public function setId(?int $id): void
```
GSadee, line 45: Similar to `CustomerBillingData`, remove the type of id not to force integers
```suggestion
    public function getId()
    {
        return $this->id;
    }

    public function setId($id): void
```

src/Entity/ShopBillingDataInterface.php
```diff
@@ -13,19 +13,35 @@
 
 namespace Sylius\RefundPlugin\Entity;
 
-interface ShopBillingDataInterface
+use Sylius\Component\Resource\Model\ResourceInterface;
+
+interface ShopBillingDataInterface extends ResourceInterface
 {
-    public function id(): ?int;
```
GSadee, line 18: Deprecate this method instead of removing it now

src/Entity/ShopBillingDataInterface.php
```diff
@@ -13,19 +13,35 @@
...
+    public function getCity(): ?string;
 
-    public function street(): ?string;
+    public function setCity(?string $city): void;
 
-    public function city(): ?string;
+    public function getPostcode(): ?string;
 
-    public function postcode(): ?string;
+    public function setPostcode(?string $postcode): void;
```
GSadee, line 46: As I see, you didn't remove these methods in `ShopBillingData` class, so why did you change them here? I would be for deprecating them instead of removing

src/Resources/config/services/generators.xml
```diff
@@ -34,6 +35,16 @@
             <argument>@SyliusRefundPlugin/Download/creditMemo.html.twig</argument>
             <argument>%sylius.refund.template.logo_file%</argument>
         </service>
+
+        <service
+                id="Sylius\RefundPlugin\Factory\ShopBillingDataFactory"
+                decorates="sylius_refund.factory.shop_billing_data"
+                decoration-priority="256"
+                public="false"
```
GSadee, line 43: Wrong indentation

UPGRADE.md
```diff
@@ -23,6 +23,23 @@
...
+
+1. The constructor of `Sylius\RefundPlugin\Generator\CreditMemoGenerator` has been changed:
+
+    ```diff
+        public function __construct(
+            LineItemsConverterInterface $lineItemsConverter,
+            LineItemsConverterInterface $shipmentLineItemsConverter,
+            TaxItemsGeneratorInterface $taxItemsGenerator,
+            CreditMemoFactoryInterface $creditMemoFactory,
+    +       ShopBillingDataFactoryInterface $shopBillingDataFactory
```
arti0090, line 37: You need to change it after rebase 😉

src/Entity/ShopBillingDataInterface.php
arti0090: This might come in handy 🙂  : https://github.com/Sylius/RefundPlugin/pull/313/files#diff-830a3324f5d10ec569eaf1c45a46c3859d022d2980c3ce2dd538ff4c5555bd14R20

arti0090: You can also check in PR I pasted how to suppress phpstan errors that will popup. See `phpstan.neon` file.

spec/Entity/ShopBillingDataSpec.php
```diff
@@ -18,48 +18,49 @@
...
         $this->shouldImplement(ShopBillingDataInterface::class);
     }
 
     public function it_has_no_id_by_default(): void
     {
-        $this->id()->shouldReturn(null);
+        $this->getId()->shouldReturn(null);
     }
 
     public function it_has_company(): void
```
arti0090, line 31: This is not a blocker for this PR but I would encourage you to add `a` or `an` articles before param.
F.e: `it_has_company` -> `it_has_a_company`

src/Generator/CreditMemoGenerator.php
```diff
@@ -117,7 +123,7 @@ private function getToAddress(?ChannelShopBillingData $channelShopBillingData):
             return null;
         }
 
-        return new ShopBillingData(
+        return $this->shopBillingDataFactory->createWithData(
```
arti0090, line 126: Maybe a method in factory called `createWithChannelShopBillingData` ? 😉 
See: https://github.com/Sylius/RefundPlugin/pull/313/files#diff-dc7a9ca1a71645aa3eb35cf2d342ab0543241c3b0173c6fe7d6ddf42a641be35R65

It won't block a merge imo - just a suggestion.

src/Entity/ShopBillingDataInterface.php
SirDomin: i got them back along with deprecation info

spec/Entity/ShopBillingDataSpec.php
SirDomin: yeah, right

src/Generator/CreditMemoGenerator.php
SirDomin: i would leave it as it is now

src/Resources/config/services/generators.xml
```diff
@@ -35,6 +37,16 @@
...
         </service>
+
+        <service
+            id="Sylius\RefundPlugin\Factory\ShopBillingDataFactory"
+            decorates="sylius_refund.factory.shop_billing_data"
+            decoration-priority="256"
+            public="false"
+        >
+            <argument type="service" id="Sylius\RefundPlugin\Factory\ShopBillingDataFactory.inner" />
+        </service>
```
Zales0123, line 48: Ok, that's why we do not specify a factory in `config.yml` 🤔 Maybe we should skip the decoration? We can do a factory + entry in resource configuration, the same way as it's done for `RefundPayment` (and it would keep an application consistency 💃 )

Zales0123 (commented):
Removal of `tests/Application/var/.gitignore` is not relevant in the scope of this PR 🖖

src/Resources/config/app/config.yml
```diff
@@ -22,6 +22,9 @@ sylius_resource:
         sylius_refund.customer_billing_data:
             classes:
                 model: Sylius\RefundPlugin\Entity\CustomerBillingData
+        sylius_refund.shop_billing_data:
+            classes:
+                model: Sylius\RefundPlugin\Entity\ShopBillingData
```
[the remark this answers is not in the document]
arti0090: @Zales0123 it makes some automagic and will lead to errors. I had some troubles when I left `factory` here. Maybe @lchrusciel could say something more as he helped me finding issue with it

src/Resources/config/services/generators.xml
arti0090: Well I leaved comment in discussion above, but thing about specify a factory was something with this.

src/Resources/config/app/config.yml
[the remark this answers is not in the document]
Zales0123: My problem is, we're actually creating a resource, so **conceptually** it's perfectly reasonable to implement a _default_ factory here, which then can be decorated in the final application 🚀 If it does not work, then wake up, Sylius Team - we have a bug to fix 🤖

src/Entity/ShopBillingDataInterface.php
```diff
@@ -13,19 +13,56 @@
...
 
-interface ShopBillingDataInterface
+use Sylius\Component\Resource\Model\ResourceInterface;
+
+interface ShopBillingDataInterface extends ResourceInterface
 {
-    public function id(): ?int;
+    public function getId();
+
+    public function setId($id): void;
```
pamil, line 22: We won't need `setId` on the interface - see https://github.com/Sylius/RefundPlugin/pull/315/files#r653493427.

src/Entity/ShopBillingDataInterface.php
```diff
@@ -13,19 +13,56 @@
 
 namespace Sylius\RefundPlugin\Entity;
 
-interface ShopBillingDataInterface
+use Sylius\Component\Resource\Model\ResourceInterface;
+
+interface ShopBillingDataInterface extends ResourceInterface
 {
-    public function id(): ?int;
+    public function getId();
```
pamil, line 20:
```suggestion
    /** @return mixed */
    public function getId();
```

phpstan.neon
```diff
@@ -19,3 +19,9 @@ parameters:
         - '/Method Sylius\\RefundPlugin\\Entity\\CustomerBillingData::setId\(\) has parameter \$id with no typehint specified./'
         - '/Method Sylius\\RefundPlugin\\Entity\\CustomerBillingDataInterface::id\(\) has no return typehint specified./'
         - '/Method Sylius\\RefundPlugin\\Entity\\CustomerBillingDataInterface::setId\(\) has parameter \$id with no typehint specified./'
+        - '/Method Sylius\\RefundPlugin\\Entity\\ShopBillingData::id\(\) has no return typehint specified./'
```
pamil, line 22: We won't need those exclusions if we explicitly define return types as mixed.

src/Entity/ShopBillingDataInterface.php
```diff
@@ -13,19 +13,56 @@
...
+
+    public function getCity(): ?string;
+
+    public function setCity(?string $city): void;
+
+    public function getPostcode(): ?string;
+
+    public function setPostcode(?string $postcode): void;
+
+    /** @deprecated this function is deprecated and will be removed in v1.0.0. Use CustomerBillingDataInterface::getId() instead */
```
pamil, line 48:
```suggestion
    /** 
     * @deprecated this function is deprecated and will be removed in v1.0.0. Use CustomerBillingDataInterface::getId() instead 
     * @see CustomerBillingDataInterface::getId
     * @return mixed
     */
```

src/Resources/config/services/generators.xml
pamil: I'd try to use a factory here, it shouldn't cause any errors in this relatively easy use-case.

pamil (dismissed):

src/Entity/ShopBillingData.php
```diff
@@ -37,23 +37,77 @@ class ShopBillingData implements ShopBillingDataInterface
...
     protected $postcode;
 
-    public function __construct(
-        ?string $company,
-        ?string $taxId,
-        ?string $countryCode,
-        ?string $street,
-        ?string $city,
-        ?string $postcode
-    ) {
```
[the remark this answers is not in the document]
SirDomin: ShopBIllingDataFactory::create new is still supported, i dont know why should we throw unsupportedMethod?

UPGRADE.md
```diff
@@ -26,6 +26,9 @@
 1. Constructor of `Sylius\RefundPlugin\Entity\CustomerBillingData` has been removed and now `CustomerBillingData` entity
    is created by `Sylius\RefundPlugin\Factory\CustomerBillingDataFactory`.
 
+1. Constructor of `Sylius\RefundPlugin\Entity\ShopBillingData` has been removed and now `ShopBillingData` entity
```
Zales0123, line 29: `Constructor of `Sylius\RefundPlugin\Entity\ShopBillingData` has been removed` nope 💃

src/Factory/ShopBillingDataFactory.php
```diff
@@ -0,0 +1,37 @@
...
+
+use Sylius\Component\Resource\Exception\UnsupportedMethodException;
+use Sylius\RefundPlugin\Entity\ShopBillingData;
+use Sylius\RefundPlugin\Entity\ShopBillingDataInterface;
+
+final class ShopBillingDataFactory implements ShopBillingDataFactoryInterface
+{
+    public function createNew(): ShopBillingDataInterface
+    {
+        throw new UnsupportedMethodException('createNew');
```
Zales0123, line 24: Exception message should be a little bit more explainable, something like: `To create ShopBillingData, you need to provide some arguments. Use ShopBillingDataFactoryInterface::createWithData instead`

src/Resources/config/app/config.yml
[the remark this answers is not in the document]
Zales0123: I suppose we're still missing this configuration?

Zales0123 (commented):
`tests/Application/var/.gitignore` is still removed and it should not be 😄

src/Factory/ShopBillingDataFactory.php
```diff
@@ -0,0 +1,46 @@
...
+
+    public function createWithData(
+        ?string $company,
+        ?string $taxId,
+        ?string $countryCode,
+        ?string $street,
+        ?string $city,
+        ?string $postcode
+    ): ShopBillingDataInterface {
+        $shopBillingData = new ShopBillingData();
```
GSadee, line 35: You should use here `className` similar to: https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/Order/Factory/OrderItemUnitFactory.php#L34 or go back to decorating the factory

src/Resources/config/app/config.yml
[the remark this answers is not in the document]
GSadee: I've searched the Sylius and there is used only in one place different factory that default one: https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/OrderBundle/DependencyInjection/Configuration.php#L110. All the others decorate the default one, so I'm not sure what is the recommended way to do. I would be probably for decorating, as it is more consistent and there is no problem with the type of created entity.

spec/Factory/ShopBillingDataFactorySpec.php
```diff
@@ -0,0 +1,49 @@
...
+declare(strict_types=1);
+
+namespace spec\Sylius\RefundPlugin\Factory;
+
+use PhpSpec\ObjectBehavior;
+use Sylius\Component\Resource\Exception\UnsupportedMethodException;
+use Sylius\RefundPlugin\Entity\ShopBillingData;
+use Sylius\RefundPlugin\Factory\ShopBillingDataFactoryInterface;
+
+class ShopBillingDataFactorySpec extends ObjectBehavior
```
Zales0123, line 21: Missing final

src/Factory/ShopBillingDataFactory.php
Zales0123: As `$className` is private property (😞) of a parent class, we should probably override a constructor and assign it to the property here as well. But yes, using this parameter is required 👌

src/Factory/ShopBillingDataFactoryInterface.php
```diff
@@ -0,0 +1,31 @@
...
+declare(strict_types=1);
+
+namespace Sylius\RefundPlugin\Factory;
+
+use Sylius\Component\Resource\Factory\FactoryInterface;
+use Sylius\RefundPlugin\Entity\ShopBillingDataInterface;
+
+interface ShopBillingDataFactoryInterface extends FactoryInterface
+{
+    public function createNew(): ShopBillingDataInterface;
```
Zales0123, line 21: I believe we don't have to define this method, as we extend the `FactoryInterface`

src/Resources/config/app/config.yml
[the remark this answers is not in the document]
Zales0123: 👍 for consistency, even though I'm sad, as it would make more sense to use this configuration as a default approach 😄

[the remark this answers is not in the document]
SirDomin: i got it back to solution with customer_billing_data

src/Factory/ShopBillingDataFactory.php
SirDomin: its decorated now

src/Factory/ShopBillingDataFactoryInterface.php
SirDomin: i reverted it to decorating

Zales0123: Still not needed, the method is in `FactoryInterface` nonetheless 😄

src/Resources/config/services/factories.xml
```diff
@@ -31,6 +31,15 @@
             <argument type="service" id="Sylius\RefundPlugin\Provider\CurrentDateTimeImmutableProviderInterface" />
         </service>
 
+        <service
+                id="Sylius\RefundPlugin\Factory\ShopBillingDataFactory"
+                decorates="sylius_refund.factory.shop_billing_data"
+                decoration-priority="256"
+                public="false"
```
Zales0123, line 38: Too big indent 🖖

src/Factory/ShopBillingDataFactoryInterface.php
SirDomin: yeah its not required to add it in this interface aswell

Zales0123 (approved):

Zales0123:
Thank you, @SirDomin! :1st_place_medal:

PR #317, opened 2021-06-18
Replace Sonata block events with Sylius template events

GSadee:


src/Resources/config/services/listeners.xml
```diff
@@ -15,15 +15,5 @@
             <argument type="service" id="Sylius\RefundPlugin\Checker\OrderRefundsListAvailabilityChecker" />
             <tag name="kernel.event_listener" event="sylius.menu.admin.order.show" method="addRefundsButton" />
         </service>
-
-        <service id="sylius_refund_plugin.block_event_listener.account.order_show" class="Sylius\Bundle\UiBundle\Block\BlockEventListener">
-            <argument>@SyliusRefundPlugin/Order/Shop/creditMemos.html.twig</argument>
-            <tag name="kernel.event_listener" event="sonata.block.event.sylius.shop.account.order.show.after_summary" method="onBlockEvent" />
-        </service>
-
-        <service id="sylius_refund_plugin.block_event_listener.order_show.credit_memos" class="Sylius\Bundle\UiBundle\Block\BlockEventListener">
```
Zales0123, line 24: Maybe we should mention it in the UPGRADE file? Someone could override these services to inject other templated (unlikely, but possible) and without removing them and switching to sylius UI events it would result in duplicated UI blocks :)

GSadee: Added 👌🏻

Zales0123 (approved):

Zales0123:

PR #321, opened 2021-06-28
Minor improvements for RefundFactory and RefundPaymentFactory

GSadee:


src/Factory/RefundPaymentFactory.php
```diff
@@ -27,6 +27,11 @@ public function __construct(string $className)
         $this->className = $className;
     }
 
+    public function createNew(): RefundPaymentInterface
+    {
+        throw new \InvalidArgumentException('This object is not default constructable.');
```
lchrusciel, line 32:
```suggestion
        throw new \InvalidArgumentException('This object is not constructable by default.');
```
or 
```suggestion
        throw new \InvalidArgumentException('This object is not constructable without any arguments.');
```
or 
```suggestion
        throw new \InvalidArgumentException('Default creation method is forbidden for this object. Use `createWithData` instead.');
```

lchrusciel (approved):

lchrusciel:

PR #323, opened 2021-07-02
Improve customization of RefundType

### LINKED ISSUE #279: Is there a way to override the RefundType without override every service using it ?

GSadee:
Fixes #279 
Replaces #311

### DIFF
--- a/src/Entity/Type/RefundEnumType.php
+++ b/src/Entity/Type/RefundEnumType.php
@@ -17,8 +17,9 @@ use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\DBAL\Types\ConversionException;
 use Doctrine\DBAL\Types\Type;
 use Sylius\RefundPlugin\Model\RefundType;
+use Sylius\RefundPlugin\Model\RefundTypeInterface;
 
-final class RefundEnumType extends Type
+class RefundEnumType extends Type
 {
     public function getName(): string
     {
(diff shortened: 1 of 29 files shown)


src/Entity/Type/RefundEnumType.php
```diff
@@ -30,31 +32,36 @@ public function getSQLDeclaration(array $fieldDeclaration, AbstractPlatform $pla
         return 'VARCHAR(256)';
     }
 
-    public function convertToPHPValue($value, AbstractPlatform $platform): RefundType
+    public function convertToPHPValue($value, AbstractPlatform $platform): RefundTypeInterface
```
AdamKasp, line 35:
```suggestion
    public function convertToPhpValue($value, AbstractPlatform $platform): RefundTypeInterface
```
this way is more camelCase :)
like `getId` instead `getID`

AdamKasp (approved):

src/Entity/Type/RefundEnumType.php
GSadee: I cannot change the name of this method, it is defined in class below

arti0090 (approved):

SirDomin (approved):

AdamKasp:
Thank you, Grzegorz! :1st_place_medal:

PR #327, opened 2021-07-26
fix for `doctrine:schema:update`

er1z:
Currently, if you install a plug-in via migrations, everything is fine.

But if you try with `doctrine:schema:update`, there's always a query that is executed, but not detected if you re-issue command:
```
ALTER TABLE sylius_refund_refund CHANGE type type VARCHAR(256) NOT NULL;
```

It's caused by lack of the comment for column. Having the `Type` definition updated solves this problem.


src/Entity/Type/RefundEnumType.php
```diff
@@ -64,4 +64,11 @@ protected function createType(string $value): RefundTypeInterface
     {
         return new RefundType($value);
     }
+
+    public function requiresSQLCommentHint(AbstractPlatform $platform)
+    {
+        return true;
+    }
+
+
```
Zales0123, line 73: Extra blank lines

src/Entity/Type/RefundEnumType.php
```diff
@@ -64,4 +64,11 @@ protected function createType(string $value): RefundTypeInterface
     {
         return new RefundType($value);
     }
+
+    public function requiresSQLCommentHint(AbstractPlatform $platform)
```
Zales0123, line 68: We should add a return type here `: bool` as well as put it before `createType` function to keep methods order

Zales0123 (commented):
To be sure that this fix work, I would add checking a `doctrine:schema:update` command into our workflow. I assume an additional step [before this one](https://github.com/Sylius/RefundPlugin/blob/master/.github/workflows/build.yml#L130) would do the trick. Could be something like:

```yaml
-
	name: Check out database schema update
	run: |
		(cd tests/Application && bin/console doctrine:database:create)
		(cd tests/Application && bin/console doctrine:schema:update)
		(cd tests/Application && bin/console doctrine:schema:validate)
		(cd tests/Application && bin/console doctrine:database:drop --force)
```

Zales0123:
The base of this pull-request was changed, you need fetch and reset your local branch 
if you want to add new commits to this pull request. **Reset before you pull, else commits
may become messed-up.**

Unless you added new commits (to this branch) locally that you did not push yet, 
execute `git fetch origin && git reset "master"` to update your local branch.
 
Feel free to ask for assistance when you get stuck :+1:

PR #328, opened 2021-07-27
fix for `doctrine:schema:update`

er1z:
Currently, if you install a plug-in via migrations, everything is fine.

But if you try with doctrine:schema:update, there's always a query that is executed, but not detected if you re-issue command:

ALTER TABLE sylius_refund_refund CHANGE type type VARCHAR(256) NOT NULL;
It's caused by lack of the comment for column. Having the Type definition updated solves this problem.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -127,6 +127,13 @@ jobs:
                 name: Install JS dependencies
                 run: (cd tests/Application && yarn install)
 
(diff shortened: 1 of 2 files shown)


Zales0123 (approved):

GSadee (approved):

GSadee:
Thanks, Przemysław! :tada:

PR #332, opened 2021-09-16
[Maintenace] Remove deprecated methods

GSadee:


arti0090 (approved):

lchrusciel (changes requested):
I would prefer to change the dock blocks that these methods will be removed in 2.0. Perhaps, we can merge this PR to master with the assumption, that the next release will be 2.0 indeed.

PR #361, opened 2022-03-24
Enable local file access in CreditMemoPdfFileGenerator

coldic3:
| Q               | A
| --------------- | -----
| Branch?         | 1.1
| Bug fix?        | yes
| New feature?    | no
| Related tickets | 

`wkhtmltopdf` disables local file access by default in the newest version (v0.12.6). This is the reason why InvoicingPlugin and RefundPlugin can’t access the logo file. The problem can be resolved with the option `enable-local-file-access` in `knp_snappy.pdf.options`. Here is the same problem on Stack Overflow: https://stackoverflow.com/questions/62315246/wkhtmltopdf-0-12-6-warning-blocked-access-to-file. I've checked it on v0.12.5 and can confirm it appears only on v0.12.6.

So now enabling local file access resolves the issue but it’s tricky and requires one more step in the installation instruction for both InvoicingPlugin and RefundPlugin. Furthermore keeping it disabled globally can be better from a security point of view. All things considered, I add this option only in `CreditMemoPdfFileGenerator`.
Please let me know if the solution is fine, then I will add a similar fix to the InvoicingPlugin :slightly_smiling_face:


Rafikooo (approved):

Ferror:
Does the `enable-local-file-access` parameter was introduced in 0.12.6 or it has been present before?

vvasiloi:
@Ferror it was introduced in 0.12.6.

vvasiloi:
@Ferror sorry, I double checked and the option is there since 2010, what changed in 0.12.6 is that the access is not allowed by default.

vvasiloi:
Instead of adding `--enable-local-file-access`, I think it's better to add `%sylius.refund.template.logo_file%` and other relevant files/dirs to the `allow` list.

coldic3:
@vvasiloi Totally agree. I've created a new generator `TwigToPdfGenerator` that wraps the knp snappy and dynamically adds allowed files to its generator. What do you think about this approach?

vvasiloi:
@coldic3 I would rather add it to the global Snappy configuration, or create a separate configuration / parameter for this service.
Having the array defined directly in the class will make it more difficult to customize in case the user wants to add more files / directories to it.

src/Resources/config/app/config.yml
```diff
@@ -2,6 +2,11 @@ parameters:
     sylius_refund.refund_enum_type: Sylius\RefundPlugin\Entity\Type\RefundEnumType
     sylius_refund.refund_type: Sylius\RefundPlugin\Model\RefundType
 
+sylius_refund:
+    pdf_generator:
+        allowed_files:
+            - '%env(default:default_logo_file:resolve:SYLIUS_REFUND_LOGO_FILE)%'
```
Ferror, line 8: Where's this env coming from?

vvasiloi: The user can define it.

src/Resources/config/app/config.yml
```diff
@@ -2,6 +2,11 @@ parameters:
     sylius_refund.refund_enum_type: Sylius\RefundPlugin\Entity\Type\RefundEnumType
     sylius_refund.refund_type: Sylius\RefundPlugin\Model\RefundType
 
+sylius_refund:
+    pdf_generator:
+        allowed_files:
```
GSadee, line 7: Why can't you just use `knp_snappy.pdf.options.allow` and allow there for the logo instead of defining the new parameter?

src/Generator/CreditMemoPdfFileGenerator.php
```diff
@@ -40,18 +36,16 @@ final class CreditMemoPdfFileGenerator implements CreditMemoPdfFileGeneratorInte
 
     public function __construct(
         RepositoryInterface $creditMemoRepository,
-        Environment $twig,
-        GeneratorInterface $pdfGenerator,
+        TwigToPdfGeneratorInterface $twigToPdfGenerator,
```
GSadee, line 39: This change is a BC break and we cannot do that this way, you should deprecate the arguments you are removing and a new one as a last and nullable argument. And use the new one only if it is not null. However, I'm not sure if this code is needed according to my other comment.

src/Resources/config/app/config.yml
coldic3: I was thinking about it and did it that way because `knp_snappy.pdf.options.allow` applies to all pdf generators that use KnpSnappy. If someone already uses KnpSnappy in their project, may want to have different allowed files for their own generator, and different for the generator coming from Sylius 🤔
But I don't know my considerations make any sense so I'm open to applying your suggestion.

coldic3: Oh, I think I've even tried it that way. I could pass only strings to `knp_snappy.pdf.options.allow` so the env variable didn't work. But I'm not sure about it right now. I'll try again and confirm it.

GSadee: Looking at the configuration suggested in [our docs](https://docs.sylius.com/en/latest/cookbook/payments/customizing-credit-memo.html#how-to-change-the-logo-in-the-credit-memo), passing env variable should work 🤔  
```
knp_snappy:
    pdf:
        options:
            allow: '%env(resolve:SYLIUS_REFUND_LOGO_FILE)%'
```

coldic3: Yep, I've got an error:
> Invalid` type for path "knp_snappy.pdf.options.allow". Expected "scalar", but got "array".
For the following configuration:
```yaml
knp_snappy:
    pdf:
        options:
            allow: '%env(default:default_logo_file:resolve:SYLIUS_REFUND_LOGO_FILE)%'
```
but
```yaml
knp_snappy:
    pdf:
        options:
            allow: '%env(SYLIUS_REFUND_LOGO_FILE)%'
```
already works, however `SYLIUS_REFUND_LOGO_FILE` may not be set.

GSadee: What about `allow: '%env(default:default_logo_file:SYLIUS_REFUND_LOGO_FILE)%'` 👀

vvasiloi: It's assigning a plugin specific variable `SYLIUS_REFUND_LOGO_FILE` to a project-wide configuration.
How will the user configure paths unrelated to the plugin? What about the invoicing plugin?
I prefer plugin-owned configuration.

GSadee: I thought that there is a possibility to define an array under the `allow` config, I didn't pay attention to: 
> Invalid` type for path "knp_snappy.pdf.options.allow". Expected "scalar", but got "array".

Now, the addition plugin configuration makes sense to me 👌🏻

Ferror (approved):

src/Generator/CreditMemoPdfFileGenerator.php
coldic3: Thanks, I removed BC and did deprecations in commit https://github.com/Sylius/RefundPlugin/pull/361/commits/22f818c861d1c10ba8481875aa8bd560d33814db. However, I worry I overdid it with that many `@trigger_error` ;d

ernestWarwas (approved):

src/Generator/CreditMemoPdfFileGenerator.php
```diff
@@ -28,30 +28,36 @@ final class CreditMemoPdfFileGenerator implements CreditMemoPdfFileGeneratorInte
...
+        string $creditMemoLogoPath,
+        ?TwigToPdfGeneratorInterface $twigToPdfGenerator = null
     ) {
         $this->creditMemoRepository = $creditMemoRepository;
         $this->twig = $twig;
         $this->pdfGenerator = $pdfGenerator;
         $this->fileLocator = $fileLocator;
         $this->template = $template;
         $this->creditMemoLogoPath = $creditMemoLogoPath;
+        $this->twigToPdfGenerator = $twigToPdfGenerator;
```
GSadee, line 58: I would be for adding here a new argument, not a `TwigToPdfGeneratorInterface`, but a new service that provides options for generating PDF file. You can then use the current logic and provide options only if the service is available. Because as I see, the logic extracted to `TwigToPdfGeneratorInterface` is mainly preparing options. Then you will need to deprecate only not using the new service.

GSadee (commented):
To check if its properly work with the proper version of wkhtmltopdf, you can install it during GitHub Actions workflow instead of using the one from tests/Application:
```
-
    name: Install wkthtmltopdf
    run: |
        sudo apt-get install -y xfonts-base xfonts-75dpi
        wget https://github.com/wkhtmltopdf/packaging/releases/download/0.12.6-1/wkhtmltox_0.12.6-1.focal_amd64.deb
        sudo dpkg -i wkhtmltox_0.12.6-1.focal_amd64.deb
```

src/Generator/CreditMemoPdfFileGenerator.php
coldic3: Ok, I've replaced `TwigToPdfGeneratorInterface` with `PdfOptionsGeneratorInterface` and will add `TwigToPdfGeneratorInterface` in the separate PR.

GSadee (approved):

GSadee:
Thank you, Kevin! :1st_place_medal:

src/Generator/PdfOptionsGenerator.php
```diff
@@ -0,0 +1,48 @@
...
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\RefundPlugin\Generator;
+
+use Symfony\Component\Config\FileLocatorInterface;
+
+final class PdfOptionsGenerator implements PdfOptionsGeneratorInterface
```
lchrusciel, line 18: I'm wondering how could be extract common logic from this and invoicing plugin

vvasiloi: To snappy maybe?! 😄 
Or yet another package in Sylius Stack... 🤷‍♂️

PR #367, opened 2022-04-20
[Maintenance] Optional pdf generator

Rafikooo:
| Q               | A
| --------------- | -----
| Branch?         | 1.2
| Bug fix?        | no
| New feature?    | yes

To not force users to install the `wkhtmltopdf` binary we introduced a configuration that allows disabling PDF generation.

```yaml
# config/packages/sylius_refund.yaml
sylius_refund:
    pdf_generator:
        enabled: false
```

Turned on:
image

Turned off:
image

Turned on:
image

Turned off:
image

Turned on:
image

Turned off:
image

And same on the customer account page


src/Action/Admin/DownloadCreditMemoAction.php
```diff
@@ -20,22 +20,21 @@
...
-    private CreditMemoPdfFileGeneratorInterface $creditMemoPdfFileGenerator;
-
-    private CreditMemoFileResponseBuilderInterface $creditMemoFileResponseBuilder;
-
     public function __construct(
-        CreditMemoPdfFileGeneratorInterface $creditMemoPdfFileGenerator,
-        CreditMemoFileResponseBuilderInterface $creditMemoFileResponseBuilder
+        private CreditMemoPdfFileGeneratorInterface $creditMemoPdfFileGenerator,
+        private CreditMemoFileResponseBuilderInterface $creditMemoFileResponseBuilder,
+        private ?string $hasEnabledPdfFileGenerator = null
```
GSadee, line 26:
```suggestion
        private ?bool $hasEnabledPdfFileGenerator = null
```
or even
```suggestion
        private bool $hasEnabledPdfFileGenerator = true
```

src/Resources/config/services.xml
```diff
@@ -10,7 +10,7 @@
 
     <parameters>
         <parameter key="default_logo_file">@SyliusRefundPlugin/Resources/assets/sylius-logo.png</parameter>
-        <parameter key="sylius.refund.template.logo_file">%env(default:default_logo_file:resolve:SYLIUS_REFUND_LOGO_FILE)%</parameter>
+        <parameter key="sylius_refund.template.logo_file">%env(default:default_logo_file:resolve:SYLIUS_REFUND_LOGO_FILE)%</parameter>
```
GSadee, line 13: You shouldn't change the parameter name as it could be used by someone in the end application, you can add a new parameter and deprecate the old one.

src/Resources/views/Order/Admin/CreditMemo/_detailsButton.html.twig
```diff
@@ -0,0 +1,7 @@
+{% import '@SyliusUi/Macro/buttons.html.twig' as buttons %}
+{{ buttons.default(
```
GSadee, line 2:
```suggestion

{{ buttons.default(
```

src/Resources/views/Order/Admin/CreditMemo/_downloadButton.html.twig
```diff
@@ -0,0 +1,8 @@
+{% import '@SyliusUi/Macro/buttons.html.twig' as buttons %}
+{{ buttons.default(
```
GSadee, line 2:
```suggestion

{{ buttons.default(
```

src/Action/Admin/DownloadCreditMemoAction.php
AdamKasp, line 26: shouldn't be boolean? In the configuration, it is configured as a boolean.
why do we need to make it nullable? If I see correctly you've configured the default value as a true

AdamKasp:
And you need to check `send` and `resend` too., `CreditMemoEmailSender` also creates the pdf and if I see it correct you don't cover this case

src/Resources/views/Order/Shop/_downloadButton.html.twig
```diff
@@ -0,0 +1,8 @@
+{% import '@SyliusUi/Macro/buttons.html.twig' as buttons %}
+
+{{ buttons.default(
+    path('sylius_refund_shop_credit_memo_download', {'id': credit_memo.id}),
+    'sylius_refund.ui.download'|trans,
```
Rafikooo, line 5: There is an inconsistency between `shop` and `admin` buttons. Here we have applied `|trans` filter. I wasn't sure which way is correct

src/Sender/CreditMemoEmailSender.php
```diff
@@ -22,36 +22,34 @@ final class CreditMemoEmailSender implements CreditMemoEmailSenderInterface
...
             self::UNITS_REFUNDED,
             [$recipient],
             ['creditMemo' => $creditMemo],
-            [$this->fileManager->realPath($filePath)]
+            $attachments ?? []
         );
 
-        $this->fileManager->remove($filePath);
+        if ($this->hasEnabledPdfFileGenerator) {
+            $this->fileManager->remove($filePath ?? '');
```
Rafikooo, line 52: Workaround because of:
https://github.com/Sylius/RefundPlugin/runs/6120425251?check_suite_focus=true#step:28:19

src/Resources/config/services.xml
Rafikooo: I wasn't sure how to accomplish it. For now, I reverted that change

src/Resources/config/app/config.yml
```diff
@@ -160,6 +161,36 @@ sylius_ui:
...
+                    priority: 50
+                breadcrumb:
+                    template: '@SyliusRefundPlugin/Order/Admin/CreditMemo/_breadcrumb.html.twig'
+                    priority: 40
+        sylius_refund.admin.order.credit_memo.right_metadata:
+            blocks:
+                download:
+                    template: '@SyliusRefundPlugin/Order/Admin/CreditMemo/_downloadButton.html.twig'
+                    enabled: '%sylius_refund.pdf_generator.enabled%'
+                    priority: 50
```
GSadee, line 193: Why did you start with the priority 50 for new blocks in new events?
```suggestion

        sylius_refund.shop.order.credit_memos.list.actions.buttons:
            blocks:
                download:
                    template: '@SyliusRefundPlugin/Order/Shop/_downloadButton.html.twig'
                    enabled: '%sylius_refund.pdf_generator.enabled%'
                    priority: 10

        sylius_refund.admin.order.credit_memos.list.actions.buttons:
            blocks:
                details:
                    template: '@SyliusRefundPlugin/Order/Admin/CreditMemo/_detailsButton.html.twig'
                    priority: 20
                download:
                    template: '@SyliusRefundPlugin/Order/Admin/CreditMemo/_downloadButton.html.twig'
                    enabled: '%sylius_refund.pdf_generator.enabled%'
                    priority: 10

        sylius_refund.admin.order.credit_memo.left_metadata:
            blocks:
                header:
                    template: '@SyliusRefundPlugin/Order/Admin/CreditMemo/_header.html.twig'
                    priority: 20
                breadcrumb:
                    template: '@SyliusRefundPlugin/Order/Admin/CreditMemo/_breadcrumb.html.twig'
                    priority: 10

        sylius_refund.admin.order.credit_memo.right_metadata:
            blocks:
                download:
                    template: '@SyliusRefundPlugin/Order/Admin/CreditMemo/_downloadButton.html.twig'
                    enabled: '%sylius_refund.pdf_generator.enabled%'
                    priority: 10
```

src/Resources/views/Order/Shop/_downloadButton.html.twig
GSadee: There is no need for `|trans` filter here, as it is used on message argument in the `default` macro function

src/Sender/CreditMemoEmailSender.php
```diff
@@ -22,36 +22,34 @@ final class CreditMemoEmailSender implements CreditMemoEmailSenderInterface
...
     ) {
-        $this->creditMemoPdfFileGenerator = $creditMemoPdfFileGenerator;
-        $this->sender = $sender;
-        $this->fileManager = $fileManager;
     }
 
     public function send(CreditMemoInterface $creditMemo, string $recipient): void
     {
-        $creditMemoPdfFile = $this->creditMemoPdfFileGenerator->generate($creditMemo->getId());
+        if ($this->hasEnabledPdfFileGenerator) {
```
GSadee, line 35: You can reverse check here and just send an email without attachment when the flag is disabled and do everything else when it is enabled

src/Sender/CreditMemoEmailSender.php
GSadee: Should be needed after minor refactor according to the above comment ☝🏻

GSadee (commented):
It would be nice to add a test for checking if everything will be ok when the flag is disabled and there is no wkhtmltopdf binary defined

src/Action/Shop/DownloadCreditMemoAction.php
```diff
@@ -22,32 +22,28 @@
...
-    private CreditMemoFileResponseBuilderInterface $creditMemoFileResponseBuilder;
-
     public function __construct(
-        CreditMemoPdfFileGeneratorInterface $creditMemoPdfFileGenerator,
-        CreditMemoCustomerRelationCheckerInterface $creditMemoCustomerRelationChecker,
-        CreditMemoFileResponseBuilderInterface $creditMemoFileResponseBuilder
+        private CreditMemoPdfFileGeneratorInterface $creditMemoPdfFileGenerator,
+        private CreditMemoCustomerRelationCheckerInterface $creditMemoCustomerRelationChecker,
+        private CreditMemoFileResponseBuilderInterface $creditMemoFileResponseBuilder,
+        private bool $hasEnabledPdfFileGenerator = true
```
AdamKasp, line 29: Do we need this default value? we have the parameter `enabled` set as true by default in configuration

src/Resources/config/app/config.yml
Rafikooo: I thought it would be nice to be able to add future blocks with a lower priority, starting from 50 gives, for instance, the possibility to add 4 new buttons after the download button. What do you think about it?

spec/Sender/CreditMemoEmailSenderSpec.php
```diff
@@ -23,25 +23,24 @@
...
 {
-    function let(
+    function it_implements_credit_memo_email_sender_interface(
         CreditMemoPdfFileGeneratorInterface $creditMemoPdfFileGenerator,
         SenderInterface $sender,
         FileManagerInterface $fileManager
     ): void {
-        $this->beConstructedWith($creditMemoPdfFileGenerator, $sender, $fileManager);
-    }
+        $this->beConstructedWith($creditMemoPdfFileGenerator, $sender, $fileManager, true);
```
lchrusciel, line 31: I would leave the default constructor definition in the "let" function and change the behaviour only for one test, but it is minor improvement

src/Resources/config/app/config.yml
GSadee: But we will still be able to add blocks in future, I would be for starting from `10` as we did on Sylius to be consistent with our approach there

lchrusciel (approved):

AdamKasp (approved):

AdamKasp:
Thanks, Rafał & Grzegorz! :tada:

PR #372, opened 2022-06-16
Inject more parameters in email template

adpeyre:
| Q               | A
| --------------- | -----
| Branch?         | 1.2 (bug fixes, improvements)
| Bug fix?        | no
| New feature?    | yes

Inject more parameters like other sylius e-mails (ex https://docs.sylius.com/en/latest/book/architecture/emails.html#shipment-confirmation)

### DIFF
--- a/src/Sender/CreditMemoEmailSender.php
+++ b/src/Sender/CreditMemoEmailSender.php
@@ -32,8 +32,15 @@ final class CreditMemoEmailSender implements CreditMemoEmailSenderInterface
 
     public function send(CreditMemoInterface $creditMemo, string $recipient): void
     {
+        $templateParameters = [
+            'creditMemo' => $creditMemo,
+            'order' => $creditMemo->getOrder(),
+            'localeCode' => $creditMemo->getLocaleCode(),
+            'channel' => $creditMemo->getChannel(),
+        ];
+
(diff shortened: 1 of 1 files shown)


GSadee:
Hi @adpeyre, thank you very much for your contribution. However, the change you propose doesn't really change anything, because the `CreditMemo` we provide to parameters already has all this information and can be used in the template. 

Because of that, I'm closing this PR 🖖🏻 Of course, feel free to open other issues/PRs or answer here, if I don't see something 😃

adpeyre:
Hi @GSadee 
Yes, I agree with you, these informations can be retrieved from the CreditMemo object but it's a problem if I want to use the Sylius Core email layout as all other (Sylius) emails. It needs the localCode and the channel variables for generating right URLs.

https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/CoreBundle/Resources/views/Email/layout.html.twig

GSadee:
@adpeyre thank for your quick answer, now, I see a small benefit of this change and it is ok for me, but you need to adjust the test: [CreditMemoEmailSenderSpec](https://github.com/Sylius/RefundPlugin/blob/1.3/spec/Sender/CreditMemoEmailSenderSpec.php). Can you do that? 😃

PR #373, opened 2022-06-29
Save credit memos on server during generation

### LINKED ISSUE #351: [Bug] PDFs are not stored permanently, sending breaks with "not found"
**SyliusRefundPlugin version affected**: v1.1.0

**Description**  
Sending a PDF file with async mailer results in an error:
`Handling "Symfony\Component\Mailer\Messenger\SendEmailMessage" failed: Path "/tmp/xyz.pdf" is not readable.`

In invoicing plugin this [has been addressed months ago](https://github.com/Sylius/InvoicingPlugin/pull/231) but not in this plugin.

As Sylius Plus requires both plugins I think this has to be fixed here too.

**Possible Solution**  
Store PDFs in `private/credit-memos`.

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | 1.3
| Bug fix?        | no
| New feature?    | yes
| Related tickets | fixes https://github.com/Sylius/RefundPlugin/issues/351

### DIFF
--- a/features/pdf_generation_enabled/saving_credit_memos_on_server_during_generation.feature
+++ b/features/pdf_generation_enabled/saving_credit_memos_on_server_during_generation.feature
@@ -15,7 +15,7 @@ Feature: Saving credit memos on server during generation
         And the order "#00000022" is already paid
         And I am logged in as an administrator
 
-    @todo @application
+    @application
     Scenario: Having credit memo saved on the server after the refund is made
-        When I refund all units of "#00000022" order with "Space money" payment
+        When I refund all units of "#00000022" order with "Space money" payment method
         Then the credit memo for "#00000022" order should be saved on the server
--- a/features/pdf_generation_enabled/saving_credit_memos_on_server_during_generation.feature
+++ b/features/pdf_generation_enabled/saving_credit_memos_on_server_during_generation.feature
@@ -1,0 +1,21 @@
+@refunds
+Feature: Saving credit memos on server during generation
+    In order to keep credit memos immutable for further usage
+    As a Store Owner
+    I want the credit memos to be saved on a server during generation
+
+    Background:
+        Given the store operates on a single channel in "United States"
+        And the store has a product "PHP T-Shirt" priced at "$10.00"
+        And the store allows shipping with "Galaxy Post"
+        And the store allows paying with "Space money"
+        And there is a customer "rick.sanchez@wubba-lubba-dub-dub.com" that placed an order "#00000022"
+        And the customer bought a single "PHP T-Shirt"
+        And the customer chose "Galaxy Post" shipping method to "United States" with "Space money" payment
+        And the order "#00000022" is already paid
+        And I am logged in as an administrator
+
+    @todo @application
+    Scenario: Having credit memo saved on the server after the refund is made
+        When I refund all units of "#00000022" order with "Space money" payment
+        Then the credit memo for "#00000022" order should be saved on the server
--- a/spec/Manager/CreditMemoFileManagerSpec.php
+++ b/spec/Manager/CreditMemoFileManagerSpec.php
@@ -13,6 +13,8 @@ declare(strict_types=1);
 
 namespace spec\Sylius\RefundPlugin\Manager;
 
+use Gaufrette\Exception\FileNotFound;
+use Gaufrette\File;
 use Gaufrette\FilesystemInterface;
 use PhpSpec\ObjectBehavior;
 use Sylius\RefundPlugin\Manager\CreditMemoFileManagerInterface;
@@ -42,5 +44,23 @@ final class CreditMemoFileManagerSpec extends ObjectBehavior
         $filesystem->delete('2018_05_000000006.pdf')->shouldBeCalled();
 
         $this->remove(new CreditMemoPdf('2018_05_000000006.pdf', 'CONTENT'));
+    }
+
+    function it_returns_credit_memo_pdf_for_given_file_name_from_filesystem(
+        FilesystemInterface $filesystem,
+        File $file,
+    ): void {
+        $file->getContent()->willReturn('CONTENT');
+        $filesystem->get('2018_05_000000006.pdf')->willReturn($file);
+
+        $this->get('2018_05_000000006.pdf')->shouldBeLike(new CreditMemoPdf('2018_05_000000006.pdf', 'CONTENT'));
+    }
+
+    function it_throws_an_exception_if_there_is_no_file_for_given_file_name_in_filesystem(
+        FilesystemInterface $filesystem,
+    ): void {
+        $filesystem->get('2018_05_000000006.pdf')->willThrow(FileNotFound::class);
+
+        $this->shouldThrow(FileNotFound::class)->during('get', ['2018_05_000000006.pdf']);
     }
 }
--- a/spec/Manager/CreditMemoFileManagerSpec.php
+++ b/spec/Manager/CreditMemoFileManagerSpec.php
@@ -1,0 +1,46 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace spec\Sylius\RefundPlugin\Manager;
+
+use Gaufrette\FilesystemInterface;
+use PhpSpec\ObjectBehavior;
+use Sylius\RefundPlugin\Manager\CreditMemoFileManagerInterface;
+use Sylius\RefundPlugin\Model\CreditMemoPdf;
+
+final class CreditMemoFileManagerSpec extends ObjectBehavior
+{
+    function let(FilesystemInterface $filesystem): void
+    {
+        $this->beConstructedWith($filesystem);
+    }
+
+    function it_implements_credit_memo_file_manager_interface(): void
+    {
+        $this->shouldImplement(CreditMemoFileManagerInterface::class);
+    }
+
+    function it_saves_credit_memo_pdf_in_given_filesystem(FilesystemInterface $filesystem): void
+    {
+        $filesystem->write('2018_05_000000006.pdf', 'CONTENT')->shouldBeCalled();
+
+        $this->save(new CreditMemoPdf('2018_05_000000006.pdf', 'CONTENT'));
+    }
+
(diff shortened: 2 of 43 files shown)


src/CommandHandler/GenerateCreditMemoHandler.php
```diff
@@ -17,31 +17,38 @@
...
-        ObjectManager $creditMemoManager,
-        MessageBusInterface $eventBus,
-        OrderRepositoryInterface $orderRepository
+        private CreditMemoGeneratorInterface $creditMemoGenerator,
+        private ObjectManager $creditMemoManager,
+        private MessageBusInterface $eventBus,
+        private OrderRepositoryInterface $orderRepository,
+        private ?CreditMemoPdfFileGeneratorInterface $creditMemoPdfFileGenerator = null,
+        private ?CreditMemoFileManagerInterface $creditMemoFileManager = null,
+        private ?bool $hasEnabledPdfFileGenerator = true,
```
AdamKasp: why do we need to have this variable nullable?

src/Manager/CreditMemoFileManager.php
```diff
@@ -0,0 +1,44 @@
...
+
+final class CreditMemoFileManager implements CreditMemoFileManagerInterface
+{
+    private FilesystemInterface $filesystem;
+
+    public function __construct(FilesystemInterface $filesystem)
+    {
+        $this->filesystem = $filesystem;
+    }
+
```
AdamKasp:
```suggestion
    public function __construct(private FilesystemInterface $filesystem)
    {
    }

```

tests/Behat/Context/Hook/CreditMemosContext.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
```
AdamKasp, line 2: missing license block

AdamKasp (approved):

features/pdf_generation_enabled/saving_credit_memos_on_server_during_generation.feature
```diff
@@ -0,0 +1,21 @@
...
+        And there is a customer "rick.sanchez@wubba-lubba-dub-dub.com" that placed an order "#00000022"
+        And the customer bought a single "PHP T-Shirt"
+        And the customer chose "Galaxy Post" shipping method to "United States" with "Space money" payment
+        And the order "#00000022" is already paid
+        And I am logged in as an administrator
+
+    @application
+    Scenario: Having credit memo saved on the server after the refund is made
+        When I refund all units of "#00000022" order with "Space money" payment method
+        Then the credit memo for "#00000022" order should be saved on the server
```
coldic3, line 21: It's a quite technical sentence. WDYT about sth like:
```suggestion
        When I refund all units of "#00000022" order with "Space money" payment method
        And I download the credit memo for "#00000022" order
        And I download this credit memo for the second time
        Then downloaded files should be exactly the same
```
It can be checked with md5 for example.

coldic3 (approved):

spec/Manager/CreditMemoFileManagerSpec.php
```diff
@@ -0,0 +1,66 @@
...
+namespace spec\Sylius\RefundPlugin\Manager;
+
+use Gaufrette\Exception\FileNotFound;
+use Gaufrette\File;
+use Gaufrette\FilesystemInterface;
+use PhpSpec\ObjectBehavior;
+use Sylius\RefundPlugin\Manager\CreditMemoFileManagerInterface;
+use Sylius\RefundPlugin\Model\CreditMemoPdf;
+
+final class CreditMemoFileManagerSpec extends ObjectBehavior
```
Zales0123, line 23: 👍 but for services operating on files I personally prefer making a PHPUnit test that checks if the file is actually saved (or we can have both of them :dancer:)

src/CommandHandler/GenerateCreditMemoHandler.php
```diff
@@ -61,9 +68,25 @@ public function __invoke(GenerateCreditMemo $command): void
...
+            return;
+        }
+
+        if (null === $this->creditMemoPdfFileGenerator || null === $this->creditMemoFileManager) {
+            return;
+        }
+
+        $creditMemoPdf = $this->creditMemoPdfFileGenerator->generate($creditMemo->getId());
+        $this->creditMemoFileManager->save($creditMemoPdf);
+    }
```
Zales0123: We're already injecting 3 new arguments into this service, also this method does not return anything and is quite independent. Maybe let's create a separate service for generating and saving pdf credit memos and inject it here? It would result in better encapsulation and having only one new argument injected here 🖖

src/Provider/CreditMemoFileProvider.php
```diff
@@ -0,0 +1,65 @@
...
+    private function provideByCreditMemo(CreditMemoInterface $creditMemo): CreditMemoPdf
+    {
+        $fileName = $this->creditMemoFileNameGenerator->generateForPdf($creditMemo);
+
+        try {
+            $pdf = $this->creditMemoFileManager->get($fileName);
+        } catch (FileNotFound) {
+            $pdf = $this->creditMemoPdfFileGenerator->generate($creditMemo->getId());
+            $this->creditMemoFileManager->save($pdf);
+        }
```
Zales0123: Two doubts:
1. Is there any possibility credit memo could not be created before it's provided? It seems as an application bug and we should be aware of it, but maybe I'm missing some scenario 🖖 
2. I'm generally not a fan of such a _hidden feature_ (hidden from the outside of the service). `provide` is a query operation and I would expect nothing is changed in the system if I call it. Maybe it would be better to throw some `CreditMemoNotFound` exception here and catch in on the higher level to react to it properly (as far as I see, this service is used for downloading and sending credit memos, so it should be handled there)

src/Sender/CreditMemoEmailSender.php
```diff
@@ -38,18 +47,33 @@ public function send(CreditMemoInterface $creditMemo, string $recipient): void
...
+            $creditMemoPdfFile = $this->creditMemoPdfFileGenerator->generate($creditMemo->getId());
+
+            $creditMemoPdfPath = $creditMemoPdfFile->filename();
+            $this->fileManager->createWithContent($creditMemoPdfPath, $creditMemoPdfFile->content());
+
+            $this->sendCreditMemo($creditMemo, $recipient, $this->fileManager->realPath($creditMemoPdfPath));
 
-        $filePath = $creditMemoPdfFile->filename();
-        $this->fileManager->createWithContent($filePath, $creditMemoPdfFile->content());
+            $this->fileManager->remove($creditMemoPdfPath);
```
Zales0123: Let's extract it to the private method for better readability

features/pdf_generation_enabled/saving_credit_memos_on_server_during_generation.feature
GSadee: It is an interesting suggestion, but the steps you've proposed should work also before this PR. I would be for adding a similar scenario in the separate PR and/or if such a problem occurs.

src/Provider/CreditMemoFileProvider.php
GSadee: 1. I'm not sure what moment are you asking about, creating credit memo or creating PDF?
2. It's interesting, because this kind of provider with a similar logic is your idea from InvoicingPlugin 😄 For me, there is a service for generating PDF file and this one is the one which is `on the higher level to react properly` if there is no PDF generated 😃 

As this feature is a new one provided here, there may be many credit memos that are created, but the PDF is not saved on the server, and because of that, this provider exists.

Zales0123: Nobody is perfect, apparently, even me xD I'm 👍 for checking the existence of the credit memo and generating it if it's not there (I was thinking about the scenario you mentioned about past credit memo 🚀). My comment is based directly on the SRP - if there is "Provider" it should **provide** not **provide and optionally save**. That's why I still think this functionality (catching an exception and saving if necessary) should be introduced in the classes that are using this provider. 

There is another thing there, [setting the full path](https://github.com/Sylius/RefundPlugin/pull/373/files#diff-f5b61b848602facab683ba301b90ff050f6ba64afa0d8c1c21106f96b0e71c2bR61) which also kinda violates CQS+SRP pattern :dancer: As far as I see, this full path is only used in [the one place](https://github.com/Sylius/RefundPlugin/pull/373/files#diff-f3995f1b446c71df2aef5749ff9cf1531b2970932e4985854360b707afb84f2eR64), so maybe it's not needed at all? Some `FullPathResolver` would (hehe) solve the issue, or maybe even solve it with the private method in the sender itself (it seems as only sender needs this full path) 💥

src/CommandHandler/GenerateCreditMemoHandler.php
GSadee: I've used here the created resolver, as it does what I need, but I have some doubts if this is a proper service, as its name could be misleading 🤔

src/Provider/CreditMemoFileProvider.php
GSadee: I've created the resolver service to extract this logic from the provider

tests/Unit/CreditMemoFileManagerTest.php
```diff
@@ -0,0 +1,77 @@
...
+use Sylius\RefundPlugin\Manager\CreditMemoFileManager;
+use Sylius\RefundPlugin\Manager\CreditMemoFileManagerInterface;
+use Sylius\RefundPlugin\Model\CreditMemoPdf;
+
+final class CreditMemoFileManagerTest extends TestCase
+{
+    /** @test */
+    function it_creates_file_in_given_filesystem(): void
+    {
+        $creditMemoFileManager = $this->prepareCreditMemoFileManager();
```
Zales0123, line 19: We could call this function in `setUp` method

tests/Unit/CreditMemoFileManagerTest.php
```diff
@@ -0,0 +1,77 @@
...
+
+        $creditMemoPdf = new CreditMemoPdf('credit-memo.pdf', 'test file content');
+        $creditMemoFileManager->save($creditMemoPdf);
+        $file = $creditMemoFileManager->get('credit-memo.pdf');
+
+        $this->assertEquals($creditMemoPdf, $file);
+
+        $this->clearTemporaryDirectory();
```
Zales0123, line 53: And this in `tearDown` :dancer:

Zales0123 (approved):

Zales0123:

PR #376, opened 2022-08-19
Added OrderItemUnit relationship to Refund entity.

kkevindev:
| Q               | A
| --------------- | -----
| Branch?         | 2.0
| Bug fix?        | no
| New feature?    | no
| Related tickets | none


beau-ottens (approved):

Roshyo:
This is not valid, as refunded_unit_id can be a shipment or an order item.
With that change, it would no longer be possible to refund a shipment.

PR #385, opened 2022-10-14
[Maintenance] Add support for Sylius 1.12 with Symfony 5.4

coldic3:
| Q               | A
| --------------- | -----
| Branch?         | 1.3
| Bug fix?        | no
| New feature?    | no
| Related tickets | 

To be merged on Sylius 1.12 release :woman_dancing:

Support for Symfony 6 will be added in a separate PR.


.github/workflows/build.yml
```diff
@@ -176,6 +176,7 @@ jobs:
             -
                 name: Validate composer.json
                 run: composer validate --ansi --strict
+                if: matrix.sylius != '1.12.x-dev'
```
jakubtobiasz: We can change 
`sylius: ["~1.11.2", "1.12.x-dev"]`
to
`sylius: ["~1.11.2", "~1.12.x-dev"]`

Then we can drop this `if` statement and we don't have to remember about this later.

tests/Application/Kernel.php
```diff
@@ -99,6 +100,18 @@ private function registerBundlesFromFile(string $bundlesFile): iterable
...
+                ['League\FlysystemBundle\FlysystemBundle' => ['all' => true]],
+                $contents,
+            );
+        } else {
+            $contents = array_merge(
+                ['Symfony\Bundle\SwiftmailerBundle\SwiftmailerBundle' => ['all' => true]],
+                $contents,
+            );
+        }
+
```
jakubtobiasz: I think putting this inside `config/bundles.php` is better as we don't have a feeling of "magically" loaded bundles.

tests/Behat/Resources/services.xml
```diff
@@ -1,6 +1,10 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 
 <container xmlns="http://symfony.com/schema/dic/services" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+    <imports>
+        <import resource="./legacy.php"/>
+    </imports>
+
```
jakubtobiasz: I'm wondering, wouldn't it be better if we put it just in `services` directory and call the file `mailer.php` or something like that 🤔.  I don't have a strong opinion, but I have mixed feelings about this.

jakubtobiasz (approved):

tests/Behat/Resources/services.xml
coldic3: Agree, it'll be better, thanks 👍️

tests/Application/Kernel.php
coldic3: I was considering it but we used to do it in the Kernel at least a few times and because of it, I would expect to find such a magic in the Kernel :P

jakubtobiasz: I'm ok with that, but I think we should talk about it in our daily meeting. I went `bundles.php` way on PPP and I would like to avoid different approaches across our plugins.

coldic3: According to our today's arrangements, I've moved it to the `bundles.php` 💃

GSadee (approved):

composer.json
```diff
@@ -23,7 +23,7 @@
         "knplabs/knp-snappy-bundle": "^1.7",
         "myclabs/php-enum": "^1.7",
         "sylius/resource-bundle": "^1.6",
-        "sylius/sylius": "~1.11.4",
+        "sylius/sylius": "~1.11.4 || ~1.12.0",
```
GSadee, line 26:
```suggestion
        "sylius/sylius": "^1.11.4",
```
but I'm not sure if your version isn't more explicit 🤔

GSadee (approved):

GSadee:
Thank you, Kevin! :1st_place_medal:

PR #389, opened 2022-11-02
Improve french translation

Nek-:
| Q               | A
| --------------- | -----
| Branch?         | 1.3 
| Bug fix?        | no
| New feature?    | no

PR #390, opened 2022-11-03
Improve french translation

Nek-:
Stands as replacement for #389 

| Q               | A
| --------------- | -----
| Branch?         | 1.2
| Bug fix?        | no
| New feature?    | no

I add a screen so @loic425 understands how wrong it is atm...

image

### DIFF
--- a/src/Resources/translations/messages.fr.yml
+++ b/src/Resources/translations/messages.fr.yml
@@ -19,7 +19,7 @@ sylius_refund:
         refund: Rembourser
         refund_all: Tout rembourser
         refund_payment_completed: Le remboursement a été effectué avec succès
-        refund_payments: Rembourser les paiements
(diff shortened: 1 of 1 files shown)


loic425:
Indeed, but that's not the translation which is wrong, but the key is.
@Zales0123 @lchrusciel @Ferror Why it is "refund_payments" instead of "refunded_payments" here?

Nek-:
I'm not sure it's a good idea to change a translation key because its content is not precise. People may override translation in their projects so keeping it like this is better.

UlrichHP:
I don't think this is the correct translation, IMHO `Remboursements` is more appropriate. It is in the same style as `Paiements` above it and is more logical because refunds are not always completed, you can have the button to finish it.

loic425:
@Nek- Not changing the key, but adding the key :dagger:

PR #391, opened 2022-11-04
Add missing csrf_token field

jakubtobiasz:
| Q               | A
| --------------- | -----
| Branch?         | 1.2
| Bug fix?        | kind of
| New feature?    | no
| Related tickets | -

In a few of our plugins, we have included this change manually, so this PR fixes this.

### DIFF
--- a/src/Action/Admin/RefundUnitsAction.php
+++ b/src/Action/Admin/RefundUnitsAction.php
@@ -23,6 +23,8 @@ use Symfony\Component\HttpFoundation\Session\Session;
 use Symfony\Component\Messenger\Exception\HandlerFailedException;
 use Symfony\Component\Messenger\MessageBusInterface;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
+use Symfony\Component\Security\Csrf\CsrfToken;
+use Symfony\Component\Security\Csrf\CsrfTokenManagerInterface;
 
 final class RefundUnitsAction
 {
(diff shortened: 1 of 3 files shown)


src/Resources/views/orderRefunds.html.twig
```diff
@@ -20,6 +20,7 @@
                     </div>
                 </div>
                 <form action="{{ path('sylius_refund_refund_units', {'orderNumber': app.request.attributes.get('orderNumber')}) }}" method="post">
+                    <input type="hidden" name="_csrf_token" value="{{ csrf_token(app.request.attributes.get('orderNumber')) }}" />
```
GSadee, line 23: Do we check this csrf token in action?

jakubtobiasz: Nope, thanks for noticing this. I don't know why we had this token in Plus, but I have added a validation now.

src/Action/Admin/RefundUnitsAction.php
```diff
@@ -36,22 +38,34 @@ final class RefundUnitsAction
...
+    private CsrfTokenManagerInterface $csrfTokenManager;
+
     public function __construct(
         MessageBusInterface $commandBus,
         Session $session,
         UrlGeneratorInterface $router,
         RefundUnitsCommandCreatorInterface $commandCreator,
-        LoggerInterface $logger
+        LoggerInterface $logger,
+        CsrfTokenManagerInterface $csrfTokenManager,
```
GSadee, line 49: As you were adding a new argument, you could use constructor property promotion here:
```suggestion
        private CsrfTokenManagerInterface $csrfTokenManager,
```

GSadee (approved):

GSadee:
Thank you, Jakub! :tada:

PR #394, opened 2022-11-17
[Maintenance] Remove wkhtmltopdf binary

coldic3:
| Q               | A
| --------------- | -----
| Branch?         | 1.3
| Bug fix?        | no
| New feature?    | no
| Related tickets | 

Let's remove the binary from the repo and install it in the GitHub Actions.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,7 +18,7 @@ jobs:
     tests:
         runs-on: ubuntu-latest
 
-        name: "Sylius ${{ matrix.sylius }}, PHP ${{ matrix.php }}, Symfony ${{ matrix.symfony }}, MySQL ${{ matrix.mysql }}, PDF: ${{ matrix.pdf }}"
+        name: "Sylius ${{ matrix.sylius }}, PHP ${{ matrix.php }}, Symfony ${{ matrix.symfony }}, MySQL ${{ matrix.mysql }}, wkhtmltopdf: ${{ matrix.wkhtmltopdf }}"
 
         strategy:
             fail-fast: false
@@ -28,7 +28,7 @@ jobs:
                 sylius: ["~1.11.4", "~1.12.0"]
                 node: ["16.x"]
                 mysql: ["5.7", "8.0"]
-                pdf: [true]
+                wkhtmltopdf: ["0.12.6-1"]
 
                 include:
                     -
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -18,7 +18,7 @@ jobs:
     tests:
         runs-on: ubuntu-latest
 
-        name: "Sylius ${{ matrix.sylius }}, PHP ${{ matrix.php }}, Symfony ${{ matrix.symfony }}, MySQL ${{ matrix.mysql }}, PDF: ${{ matrix.pdf }}"
+        name: "Sylius ${{ matrix.sylius }}, PHP ${{ matrix.php }}, Symfony ${{ matrix.symfony }}, MySQL ${{ matrix.mysql }}, wkhtmltox: ${{ matrix.wkhtmltox }}"
```
Rafikooo: Why `wkhtmltox` instead of `whhtmltopdf`?

coldic3: I can't find how to install just a `whhtmltopdf`. It seems I have to install `wkhtmltox` that contains `wkhtmltoimage` also.

coldic3: After our internal discussion, I decided to rename it to `wkhtmltopdf` 👍️

.github/workflows/build.yml
```diff
@@ -18,7 +18,7 @@ jobs:
     tests:
         runs-on: ubuntu-latest
 
-        name: "Sylius ${{ matrix.sylius }}, PHP ${{ matrix.php }}, Symfony ${{ matrix.symfony }}, MySQL ${{ matrix.mysql }}, PDF: ${{ matrix.pdf }}"
+        name: "Sylius ${{ matrix.sylius }}, PHP ${{ matrix.php }}, Symfony ${{ matrix.symfony }}, MySQL ${{ matrix.mysql }}, wkhtmltopdf: ${{ matrix.wkhtmltopdf }}"
```
GSadee, line 21: I wonder if it wouldn't be better to leave it as it was and hardcode the version of wkhtmltopdf in the proper step. IMO, this information is not important in the name of the build and in the matrix with supported versions 🤔

coldic3: The version cannot be hardcoded because of the cache. If the version is different, the cache will be invalidated. However, I can move the wkhtml version to env variables instead.

coldic3: On the other hand, this is the version of wkhtml we support and test so IMO having it in the matrix is ok. It's just a different info in the build name (exact version instead of `true` :stuck_out_tongue:)

GSadee: Ok, you've convinced me 👌🏻

GSadee:
Thanks, Kevin! :tada:

PR #396, opened 2023-02-01
`SyliusTemplateEvents` introduced on the `Refund` view

Rafikooo:
| Q               | A
| --------------- | -----
| Branch?         | 1.4
| Bug fix?        | no
| New feature?    | no

These events were introduced to ease `Plus` installation and to remove redundant templates

### DIFF
--- a/src/Resources/config/app/events.yaml
+++ b/src/Resources/config/app/events.yaml
@@ -55,3 +55,90 @@ sylius_ui:
                     template: '@SyliusRefundPlugin/Order/Admin/CreditMemo/_downloadButton.html.twig'
                     enabled: '%sylius_refund.pdf_generator.enabled%'
                     priority: 10
+
+        sylius_refund.admin.order.refund.buttons:
+            blocks:
+                clear_all:
+                    template: '@SyliusRefundPlugin/Order/Admin/Refund/Button/_clearAll.html.twig'
+                    priority: 20
+                refund_all:
+                    template: '@SyliusRefundPlugin/Order/Admin/Refund/Button/_refundAll.html.twig'
+                    priority: 10
+
+        sylius_refund.admin.order.refund:
+            blocks:
+                bulk_buttons:
+                    template: '@SyliusRefundPlugin/Order/Admin/Refund/_buttons.html.twig'
+                    priority: 20
+                form:
+                    template: '@SyliusRefundPlugin/Order/Admin/Refund/_form.html.twig'
+                    priority: 10
+
+        sylius_refund.admin.order.refund.form:
+            blocks:
+                table:
+                    template: '@SyliusRefundPlugin/Order/Admin/Refund/Form/Table/_table.html.twig'
+                    priority: 30
+                payment_method:
+                    template: '@SyliusRefundPlugin/_paymentMethod.html.twig'
+                    priority: 20
+                footer:
+                    template: '@SyliusRefundPlugin/_footer.html.twig'
+                    priority: 10
+
+        sylius_refund.admin.order.refund.form.table.header:
+            blocks:
+                item_info:
+                    template: '@SyliusRefundPlugin/Order/Admin/Refund/Form/Table/Header/_itemInfo.html.twig'
+                    priority: 40
+                total:
+                    template: '@SyliusRefundPlugin/Order/Admin/Refund/Form/Table/Header/_total.html.twig'
+                    priority: 30
+                refund_value:
+                    template: '@SyliusRefundPlugin/Order/Admin/Refund/Form/Table/Header/_refundValue.html.twig'
+                    priority: 20
+                blank:
+                    template: '@SyliusRefundPlugin/Order/Admin/Refund/Form/Table/Header/_blank.html.twig'
+                    priority: 10
+
+        sylius_refund.admin.order.refund.form.table.body:
+            blocks:
+                items:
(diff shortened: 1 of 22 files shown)


coldic3 (approved):

jakubtobiasz (commented):
Why some files are empty?

src/Resources/config/app/events.yaml
```diff
@@ -55,3 +55,69 @@ sylius_ui:
                     template: '@SyliusRefundPlugin/Order/Admin/CreditMemo/_downloadButton.html.twig'
                     enabled: '%sylius_refund.pdf_generator.enabled%'
                     priority: 10
+
+        sylius_refund.order.admin.refund.body:
```
GSadee:
```suggestion
        sylius_refund.admin.order.refund.body:
```

GSadee: And in other events below 👇🏻

src/Resources/views/Order/Admin/Refund/Body/_bulkButtons.html.twig
```diff
@@ -0,0 +1,7 @@
+<div class="ui right aligned grid">
+    <div class="sixteen wide column">
+        <button data-refund-clear type="button" class="ui button" {{ disableButton }}>{{ 'sylius_refund.ui.clear_refunds'|trans }}</button>
+        <button data-refund-all type="button" class="ui button primary" {{ disableButton }}>{{ 'sylius_refund.ui.refund_all'|trans }}</button>
```
GSadee: I'm not sure if this makes sense and if it's not overengineering, but there could be one event here and these buttons would be added by blocks

src/Resources/views/Order/Admin/Refund/Body/_form.html.twig
```diff
@@ -0,0 +1,20 @@
+<form action="{{ path('sylius_refund_refund_units', {'orderNumber': app.request.attributes.get('orderNumber')}) }}" method="post">
+    <input type="hidden" name="_csrf_token" value="{{ csrf_token(app.request.attributes.get('orderNumber')) }}" />
+    <table id="refunds" class="ui compact celled table">
+        <thead>
+        <tr>
+            {{ sylius_template_event('sylius_refund.order.admin.refund.table.header', _context) }}
+        </tr>
+        </thead>
+        <tbody>
+        {{ sylius_template_event('sylius_refund.order.admin.refund.table.row', _context) }}
```
GSadee:
```suggestion
        {{ sylius_template_event('sylius_refund.order.admin.refund.table.body', _context) }}
```
or
```suggestion
        {{ sylius_template_event('sylius_refund.order.admin.refund.table.content', _context) }}
```

src/Resources/views/Order/Admin/Refund/Body/_form.html.twig
```diff
@@ -0,0 +1,20 @@
...
+        </tbody>
+    </table>
+    {% include '@SyliusRefundPlugin/_paymentMethod.html.twig' %}
```
GSadee: In order not to have to include this template here, you could create a single template event `...form.content` with blocks for table, payment method and footer

src/Resources/views/Order/Admin/Refund/Body/_form.html.twig
```diff
@@ -0,0 +1,20 @@
...
+    <div class="ui stackable grid">
+        <div class="two column row">
+            {{ sylius_template_event('sylius_refund.admin.order.refund.footer', _context) }}
```
GSadee: According to the comment above:
```suggestion
            {{ sylius_template_event('sylius_refund.admin.order.refund.footer.content', _context) }}
```

GSadee (commented):
Changing the name of a template is a BC break, the same is about removing the existing templates, so you need to revert these changes, ref. https://docs.sylius.com/en/1.12/book/organization/backward-compatibility-promise.html?highlight=backward#templates

Rafikooo:
> Changing the name of a template is a BC break, the same is about removing the existing templates, so you need to revert these changes, ref. https://docs.sylius.com/en/1.12/book/organization/backward-compatibility-promise.html?highlight=backward#templates

I forget to check that. All deleted/renamed/moved files reverted 🖖

src/Resources/views/Order/Admin/Refund/Body/_bulkButtons.html.twig
Rafikooo: IMO good point, it opens possibility to more buttons 😄

src/Resources/views/Order/Admin/Refund/Body/_form.html.twig
Rafikooo: I've decided not to use `content` suffix as it's quite generic and could be added to every event. In fact, each block could be treated as some kind of content. Wdyt?

src/Resources/views/Order/Admin/Refund/Body/_form.html.twig
Rafikooo: Commented above

src/Resources/views/Order/Admin/Refund/Body/_form.html.twig
GSadee: It makes sense, however, we've used this keyword in template events in Sylius, but for now, there is no rule or any convention about the naming of events

GSadee (approved):

GSadee:
Thank you, Rafał! :1st_place_medal:

PR #404, opened 2023-03-08
Units refunded flow improvements

coldic3:
| Q               | A
| --------------- | -----
| Branch?         | 1.4
| Bug fix?        | no
| New feature?    | no
| Related tickets | 
| Documentation | https://github.com/Sylius/Sylius/pull/14853

The PR gives developers better extendability so you can implement your own `UnitRefund` model.

Example of `OrderItemRefund` implementation that can be done thanks to this PR: https://github.com/coldic3/Sylius-Standard/commit/7f29587827f4a7c58c81a0abca74c142a4e2aa7f


src/Creator/RefundUnitsCommandCreatorInterface.php
```diff
@@ -13,10 +13,9 @@
 
 namespace Sylius\RefundPlugin\Creator;
 
-use Sylius\RefundPlugin\Command\RefundUnits;
 use Symfony\Component\HttpFoundation\Request;
 
 interface RefundUnitsCommandCreatorInterface
 {
-    public function fromRequest(Request $request): RefundUnits;
+    public function fromRequest(Request $request): object;
```
coldic3: Because of that change, `RefundUnitsCommandCreatorInterface` might be eg. `RequestCommandCreatorInterface` as it might be implemented for any command. WDYT?

UPGRADE-2.0.md
```diff
@@ -19,3 +19,170 @@
...
+    ```
+
+13. The constructor of the `Sylius\RefundPlugin\Refunder\OrderItemUnitsRefunder` has been changed:
+
+    ```diff
+        public function __construct(
+    -       private RefundCreatorInterface $refundCreator,
+    -       private MessageBusInterface $eventBus,
+    +       private RefundCreatorInterface $refundCreator,
+    +       private MessageBusInterface $eventBus,
```
lchrusciel:
```suggestion
           private RefundCreatorInterface $refundCreator,
           private MessageBusInterface $eventBus,
           private RefundCreatorInterface $refundCreator,
           private MessageBusInterface $eventBus,
```

UPGRADE-2.0.md
```diff
@@ -19,3 +19,170 @@
                 sample_resource:
                     ...
     ```
+
+2. The constructor of the `Sylius\RefundPlugin\Command\GenerateCreditMemo` command has been changed:
```
lchrusciel: We should mention the removal of the `shipments` method.

UPGRADE-2.0.md
```diff
@@ -19,3 +19,170 @@
...
+   -       private array $shipments,      
+           private string $comment,
+       ) {
+           // ...    
+       }
+   ```
+
+   and `Sylius\RefundPlugin\Command\GenerateCreditMemo::shipments` method has been removed.
+
+3. The constructor of the `Sylius\RefundPlugin\Command\RefundUnits` command has been changed:
```
lchrusciel: We should mention the removal of the `shipments` method.

src/Filter/UnitRefundFilter.php
```diff
@@ -0,0 +1,28 @@
...
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\RefundPlugin\Filter;
+
+use Sylius\RefundPlugin\Model\UnitRefundInterface;
+
+class UnitRefundFilter implements UnitRefundFilterInterface
```
lchrusciel:
```suggestion
final class UnitRefundFilter implements UnitRefundFilterInterface
```

src/Filter/UnitRefundFilterInterface.php
```diff
@@ -0,0 +1,29 @@
...
+namespace Sylius\RefundPlugin\Filter;
+
+use Sylius\RefundPlugin\Model\UnitRefundInterface;
+
+interface UnitRefundFilterInterface
+{
+    /**
+     * @template T of UnitRefundInterface
+     *
+     * @param T[] $units
```
lchrusciel: As in the first argument, we will receive "dirty" list of all types of units, but we are saying that it will be `T`. With the current declaration, we are saying that on input, there are the following two arguments, e.g. list of `ShipmentRefund` and the FQCN of ShipmentRefund, and as output, we are returning `ShipmentRefund` again. What is filtrated then? I suppose we should mean, that there will be different types of refunds as the first argument, but the output list will return the classes that are of type `unitRefundClass`

```suggestion
     * @param UnitRefundInterface[] $units
```

src/Filter/UnitRefundFilter.php
```diff
@@ -0,0 +1,28 @@
...
+{
+    public function filterUnitRefunds(array $units, string $unitRefundClass): array
+    {
+        return array_values(
+            array_filter($units, function ($units) use ($unitRefundClass): bool {
+                return $units instanceof UnitRefundInterface && $units instanceof $unitRefundClass;
```
lchrusciel: Do we even expect `units` to be something different then UnitRefundInterface? What would you say, that we finalize either a custom collection class or just trigger an exception if `$units instanceof UnitRefundInterface` is not true?

src/Generator/CreditMemoGeneratorInterface.php
```diff
@@ -22,7 +22,6 @@ public function generate(
         OrderInterface $order,
         int $total,
         array $units,
-        array $shipments,
```
lchrusciel, line 25: we are missing this from upgrade file

lchrusciel (commented):
The main question remains, do we want to provide a forward compatible layer for refund to ease migration - I would say we should, as it seems quite straightforward.

In addition, @coldic3, what would you say to give it a try: https://github.com/Roave/BackwardCompatibilityCheck and see if they will report anything new on this PR?

UPGRADE-2.0.md
coldic3: It is mentioned right below :D

https://github.com/Sylius/RefundPlugin/pull/404/files#diff-57a2d391e31e0d1c39e9a12a231e9e3cb7e2554b7faa11c80ab43b7ce889c631R40

UPGRADE-2.0.md
coldic3: same here :P

src/Command/GenerateCreditMemo.php
```diff
@@ -13,34 +13,35 @@
...
 
-    private int $total;
+    public function __construct(
+        private string $orderNumber,
+        private int $total,
+        /** @var array|UnitRefundInterface[] */
+        private array $units,
+        private string|array $comment,
+    ) {
+        $args = func_get_args();
```
lchrusciel, line 31: What about something like that?
```suggestion
    public function __construct(
        private string $orderNumber,
        private int $total,
        /** @var array|UnitRefundInterface[] */
        private array $units,
        private string|array $shipmentsOrComment,
        ?string $comment = null,
    ) {
        if (is_array($shipmentsOrComment)) { // trigger deprecation }
        if (null !== $comment)) { // trigger deprecation }
        if (null !== $comment && is_string($shipmentsOrComment)) { // trigger deprecation }
```
I'm still considering how much we should support named arguments tho... but this could be handled with error handling and adding null to `shipmentsOrComment`

UPGRADE-1.4.md
```diff
@@ -0,0 +1,178 @@
+### UPGRADE FROM 1.X TO 2.0
```
lchrusciel: The file name and this header does not correspond well together. Let's rename it to 2.0

src/Converter/RefundUnitsConverterInterface.php
```diff
@@ -13,13 +13,16 @@
...
     /**
-     * @return UnitRefundInterface[]
+     * @template T of UnitRefundInterface
+     *
+     * @param class-string<T> $unitRefundClass
+     *
+     * @return T[]
      */
-    public function convert(array $units, RefundTypeInterface $refundType, string $unitRefundClass): array;
+    public function convert(array $units, string $unitRefundClass): array;
```
lchrusciel: this is should be reverted as well

src/Creator/RequestCommandCreatorInterface.php
```diff
@@ -13,10 +13,11 @@
...
 use Symfony\Component\HttpFoundation\Request;
 
-interface RefundUnitsCommandCreatorInterface
+interface RequestCommandCreatorInterface
 {
-    public function fromRequest(Request $request): RefundUnits;
+    public function fromRequest(Request $request): object;
 }
+
+class_alias(RequestCommandCreatorInterface::class, 'Sylius\RefundPlugin\Creator\RefundUnitsCommandCreatorInterface');
```
lchrusciel: nice 👌 

but If I remember correctly the previously existing class needs to stay where it was, to satisfy static analysers etc.

src/Filter/UnitRefundFilterInterface.php
```diff
@@ -0,0 +1,27 @@
...
+
+use Sylius\RefundPlugin\Model\UnitRefundInterface;
+
+interface UnitRefundFilterInterface
+{
+    /**
+     * @param UnitRefundInterface[] $units
+     * @param class-string<UnitRefundInterface> $unitRefundClass
+     *
+     * @return UnitRefundInterface[]
```
lchrusciel, line 24:
```suggestion
     * @template T of UnitRefundInterface
     *
     * @param UnitRefundInterface[] $units
     * @param class-string<T> $unitRefundClass
     *
     * @return T[]
```

src/Model/UnitRefundInterface.php
```diff
@@ -18,4 +18,6 @@ interface UnitRefundInterface
     public function id(): int;
 
     public function total(): int;
+
+    public static function type(): RefundType;
```
lchrusciel: This should be added in a comment to interface

src/Filter/UnitRefundFilterInterface.php
coldic3: phpstan has a problem with that
image

src/Command/GenerateCreditMemo.php
coldic3: How should it work? I see 4th and 6th arguments have the same name. Another thing is that In the previous implementation, there were 5 args and now we introduce another but the last one should never be used 🤔

UPGRADE-1.4.md
coldic3: In 2.0 we drop deprecations only so the file name should stay but the header needs to be changed. Am I right?

src/Command/GenerateCreditMemo.php
```diff
@@ -13,34 +13,35 @@
...
+    private array $shipments = [];
 
-    private int $total;
+    public function __construct(
+        private string $orderNumber,
+        private int $total,
+        /** @var array|UnitRefundInterface[] */
+        private array $units,
+        private string|array $comment,
+    ) {
```
lchrusciel, line 30:
```suggestion
    public function __construct(
        private string $orderNumber,
        private int $total,
        /** @var array|UnitRefundInterface[] */
        private array $units,
        private string|array $comment,
        /** additional argument */
    ) {
```

src/Converter/LineItemsConverterInterface.php
```diff
@@ -13,15 +13,9 @@
...
 namespace Sylius\RefundPlugin\Converter;
 
-use Sylius\RefundPlugin\Entity\LineItemInterface;
-use Sylius\RefundPlugin\Model\UnitRefundInterface;
-
-interface LineItemsConverterInterface
+/**
+ * @deprecated since 1.4, to be removed in 2.0, use Sylius\RefundPlugin\Converter\LineItem\LineItemsConverterInterface instead
+ */
+interface LineItemsConverterInterface extends LineItem\LineItemsConverterInterface
```
lchrusciel, line 19: We should anyway implement this method in classes to ensure BC.

src/Filter/UnitRefundFilter.php
```diff
@@ -0,0 +1,31 @@
...
+use Sylius\RefundPlugin\Model\UnitRefundInterface;
+use Webmozart\Assert\Assert;
+
+final class UnitRefundFilter implements UnitRefundFilterInterface
+{
+    public function filterUnitRefunds(array $units, string $unitRefundClass): array
+    {
+        return array_values(
+            array_filter($units, function ($unit) use ($unitRefundClass): bool {
+                Assert::isInstanceOf($unit, UnitRefundInterface::class);
```
lchrusciel:
```suggestion
            array_filter($units, function (UnitRefundInterface $unit) use ($unitRefundClass): bool {
```

lchrusciel (approved):

lchrusciel:
Thank you, @coldic3!

PR #407, opened 2023-03-21
Validate if refunded units belong to an order

coldic3:
| Q               | A
| --------------- | -----
| Branch?         | 1.4
| Bug fix?        | no
| New feature?    | no
| Related tickets |


src/Doctrine/ORM/UnitRefundBelongsToOrderQuery.php
```diff
@@ -0,0 +1,59 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\RefundPlugin\Doctrine\ORM;
+
+use Sylius\Component\Core\Repository\OrderItemUnitRepositoryInterface;
+use Sylius\Component\Resource\Repository\RepositoryInterface;
+
+final class UnitRefundBelongsToOrderQuery implements UnitRefundBelongsToOrderQueryInterface
```
lchrusciel: I want to discuss the naming of this class. Querying is just a part of the logic here. The equally important part is related to checking the values. I'm proposing following structure
```suggestion
final class DbBasedOrderItemUnitRefundBelongingToOrderChecker implements UnitRefundBelongingToOrderCheckerInterface
```

such class could have `check` or `areRefundsBelongingToOrder()` method

Which could then use `CountOrderItemUnitRefundBelongingToOrderQuery` with just `__invoke` or `query` method. 

As you can see, I'm also discussing the coupling of order items and shipment refund units placement in one class. It is because I don't see a good, DX-oriented way how I would add another checker that would take OrderItemRefunds into the process.

Therefore, I'm asking for:
1. decoupling of order item unit and shipment checking
2. decoupling quering logic from checking
3. allow for easier injection of new checkers for end users

src/Doctrine/ORM/CountOrderItemUnitRefundsBelongingToOrderQuery.php
```diff
@@ -0,0 +1,38 @@
...
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\RefundPlugin\Doctrine\ORM;
+
+use Sylius\Component\Core\Repository\OrderItemUnitRepositoryInterface;
+
+class CountOrderItemUnitRefundsBelongingToOrderQuery implements CountRefundsBelongingToOrderQueryInterface
```
lchrusciel: as query services with one method only, we could finalize it. There is nothing we can change in this class
```suggestion
final class CountOrderItemUnitRefundsBelongingToOrderQuery implements CountRefundsBelongingToOrderQueryInterface
```

src/Doctrine/ORM/CountShipmentRefundsBelongingToOrderQuery.php
```diff
@@ -0,0 +1,37 @@
...
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\RefundPlugin\Doctrine\ORM;
+
+use Sylius\Component\Resource\Repository\RepositoryInterface;
+
+class CountShipmentRefundsBelongingToOrderQuery implements CountRefundsBelongingToOrderQueryInterface
```
lchrusciel:
```suggestion
final class CountShipmentRefundsBelongingToOrderQuery implements CountRefundsBelongingToOrderQueryInterface
```

lchrusciel (approved):

lchrusciel:
Thank you, @coldic3!

PR #436, opened 2024-11-08
[Maintenance] Update legacy package.json in test Application

mpysiak:
| Q               | A
| --------------- | -----
| Branch?         | 1.6
| Bug fix?        | no
| New feature?    | no
| Related tickets | n/a

Also bump CI config to remove warnings
Also bump Sylius to 1.14

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,27 +29,35 @@ jobs:
             matrix:
                 php: ["8.1", "8.2"]
                 symfony: ["^5.4.21", "^6.4"]
-                sylius: ["~1.13.0", "v1.14.0-beta.1"]
+                sylius: ["~1.13.0", "1.14.0-rc.1"]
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -29,27 +29,35 @@ jobs:
             matrix:
                 php: ["8.1", "8.2"]
```
Rafikooo, line 30: In the InvoicingPlugin (https://github.com/Sylius/InvoicingPlugin/blob/1.0/.github/workflows/build.yml#L22), we kept only PHP 8.3 in the builds, as well as we did in the PayPalPlugin (https://github.com/Sylius/PayPalPlugin/blob/1.7/.github/workflows/build.yml#L22)

Rafikooo (approved):

GSadee (approved):

PR #501, opened 2026-07-21
Config/xml to php conversion

rust-le:
| Q               | A
| --------------- | -----
| Branch?         | 2.0
| Bug fix?        | no
| New feature?    | no
| Related tickets | https://github.com/GromNaN/symfony-config-xml-to-php


crydotsnake (changes requested):
Would be cool if we could cleanup all the imports in the service definition classes with `use` statements instead of using `\Sylius\RefundPlugin\Action\Admin\OrderRefundsListAction::class` etc 🙂

